PR #56, opened 2024-02-26
Feature/add json schema

MatsJohansen87:


patrickskowronekdkfz (commented):
Could this happen on compile time or before we start the `npm run dev`? This would save us some troubleshooting and compute time.

package.json
```diff
@@ -30,9 +30,9 @@
     "start": "npm run dev -s",
     "dev": "vite --config vite.demo.config.ts",
     "build": "vite build",
-    "build:demo": "vite build --config vite.demo.config.ts",
+    "build:demo": "node options_tester.cjs && vite build --config vite.demo.config.ts",
```
MatsJohansen87, line 33: should this not also go on "build"?

patrickskowronekdkfz: No, build is for the npm package. The npm packages has no options file, therefore it would not make much sense. However, this options_tester should be a of the template

patrickskowronekdkfz (approved):

PR #57, opened 2024-02-28
Add info button to individual chips and display only that child's message in dialogue

1000daysofcode:


packages/lib/src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -3,18 +3,26 @@
...
     import { iconStore } from "../../stores/icons";
-    import { getHumanReadableQuery } from "../../stores/negotiate";
 
     export let message: string[] = [];
     export let noQueryMessage: string = "Search for all results";
     export let showQuery: boolean = false;
+    
     export let infoIconUrl: string | null = null;
+    export let onlyChildInfo: boolean = false;
+    export let queryItem: QueryItem;
```
MatsJohansen87, line 25: Here we should add `undefined` as a type and assign `undefined`, otherwise we get a warning, that it has an unused export variable, when it's not set.

packages/lib/src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -33,28 +41,45 @@
...
-            message =
-                getHumanReadableQuery().length > 0
-                    ? [getHumanReadableQuery()]
-                    : [noQueryMessage];
+             if (onlyChildInfo) {
+                let childMessage =
+                    buildHumanReadableRecursively(
+                    returnNestedValues(queryItem), "");
+                message = childMessage.length > 0
+                    ? childMessage : noQueryMessage;
```
MatsJohansen87, line 51: `childMessage` and `noQueryMessage` should be in an array. That way don't need the conditional in line 76.

packages/lib/src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -33,28 +41,45 @@
...
     };
 
-    const displayQueryInfo = (): void => {
+    const displayQueryInfo = (queryItem?: AstBottomLayerValue): void => {
         if (showQuery) {
-            message =
-                getHumanReadableQuery().length > 0
-                    ? [getHumanReadableQuery()]
-                    : [noQueryMessage];
+             if (onlyChildInfo) {
```
MatsJohansen87, line 46: As `queryItem` can be undefined we should check for it to not be undefined

packages/lib/src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -33,28 +41,45 @@
...
+                let childMessage =
+                    buildHumanReadableRecursively(
+                    returnNestedValues(queryItem), "");
```
MatsJohansen87, line 49: `returnNestedValues()` can also return null, that's why typescript is complaining here. In our case it's not expected to be null, so we can cast with `as AstElement`

PR #67, opened 2024-03-19
Switch to Server Sent Events based Querying

torbrenner:
This pull request replaces the default Spot implementation with the Rust Spot implementation, which queries the sites using server send envents. Additionally the deployment files are updated to deploy the Rust based spot implementation.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -20,15 +20,13 @@ services:
 
   spot:
     # NOTE: This will be replaced by https://github.com/samply/spot soon.
-    image: docker.verbis.dkfz.de/ccp-private/central-spot
+    image: samply/rustyspot:refactor-evenMore
     ports:
       - 8080:8080
     environment:
       BEAM_SECRET: "${LOCAL_BEAM_SECRET}"
-      BEAM_URL: http://beam-proxy:8081
-      BEAM_PROXY_ID: ${LOCAL_BEAM_ID}
-      BEAM_BROKER_ID: ${BROKER_HOST}
-      BEAM_APP_ID: "focus"
+      BEAM_PROXY_URL: http://beam-proxy:8081
+      BEAM_APP_ID: "focus.${LOCAL_BEAM_ID}.${BROKER_HOST}"
     depends_on:
       - "beam-proxy"
     profiles:
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -19,8 +19,7 @@ services:
       - "default"
 
   spot:
-    # NOTE: This will be replaced by https://github.com/samply/spot soon.
-    image: samply/rustyspot:refactor-evenMore
+    image: samply/rustyspot:main
     ports:
       - 8080:8080
     environment:
--- a/packages/lib/src/classes/spot.ts
+++ b/packages/lib/src/classes/spot.ts
@@ -16,18 +16,18 @@ type BeamResult = {
     to: string[];
 };
 
+/**
+ * Implements requests to multiple targets through the middleware spot (see: https://github.com/samply/spot).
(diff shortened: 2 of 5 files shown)


docker-compose.dev.yml
```diff
@@ -20,15 +20,13 @@ services:
 
   spot:
     # NOTE: This will be replaced by https://github.com/samply/spot soon.
-    image: docker.verbis.dkfz.de/ccp-private/central-spot
+    image: samply/rustyspot:refactor-evenMore
```
Threated, line 23: This should be main too right?
```suggestion
    image: samply/rustyspot:main
```

Threated (changes requested):

packages/lib/src/classes/spot.ts
```diff
@@ -36,87 +36,59 @@ export class Spot {
...
-                if (!beamResponses.ok) {
-                    const error: string = await beamResponses.text();
-                    console.debug(
-                        `Received ${beamResponses.status} with message ${error}`,
-                    );
-                    throw new Error(
-                        `Error then retrieving responses from Beam. Abborting requests ...`,
-                    );
-                }
+            console.log(`Created new Beam Task with id ${this.currentTask}`)
```
patrickskowronekdkfz, line 65: I would use info instead of log!

packages/lib/src/classes/spot.ts
```diff
@@ -36,87 +36,59 @@ export class Spot {
...
+                    store.set(site, { status: status, data: body })
+                    return store;
+                })
+            })
 
-                    // if the site is already in the store and the status is claimed, don't update the store
-                    if (this.storeCache.get(site)?.status === status) return;
+            // read error events from beam
+            eventSource.addEventListener("error", (message) => {
+                console.log(`Beam returned error ${message}`)
```
patrickskowronekdkfz, line 83: I would use error instead of log!

patrickskowronekdkfz, line 83: If beam encounters an error, maybe we should inform the user. I'm pretty sure we should do something like this, but not in this PR.

packages/lib/src/classes/spot.ts
```diff
@@ -36,87 +36,59 @@ export class Spot {
...
-                        },
-                    );
-                }
-
-                responseCount = beamResponseData.length;
-                const realResponseCount = beamResponseData.filter(
-                    (response) => response.status !== "claimed",
-                ).length;
+            // event source in javascript throws an error then the event source is closed by backend
+            eventSource.onerror = () => {
```
patrickskowronekdkfz, line 88: Maybe add a console.error here

patrickskowronekdkfz (changes requested):
The PR looks good so far, however we should invest time into correct error handling

docker-compose.dev.yml
torbrenner: change applied

patrickskowronekdkfz (approved):

packages/lib/src/classes/spot.ts
torbrenner: Yes, i think that is a good idea. Maybe we should open an issue for this. I'think we should just provide an api for the applications to listen for such events, as each application using lens will maybe have their own way of handling it (e.g. showing the notification in the top of the application or as a box in the right bottom corner)

packages/lib/src/classes/spot.ts
torbrenner: An error log is not the right choice here. Suprisingly, the eventSource.onerror event is thrown after the backend closes the EventStream, so actually after the beam proxy received answers from all sites. I added an info log instead.

Threated (approved):

PR #95, opened 2024-05-16
Feature/make backends configurable

MatsJohansen87:
### Description
add possibility to configure multiple backends
- add internal spot and blaze support to be configured via options.json
- add APIs to get and set different stores

### Motivation and Context
It should be possible to connect multiple backends with different query languages. An out of the box support for blaze and spot with CQL is needed as it is the most common use. But the library user may want to connect their own backend(s) (in addition).

### How Has This Been Tested?
Implemented a spot in the development project and queried it simultaneously with the spot provided from Lens.

### Screenshots (if appropriate):

---


packages/demo/public/options.json
```diff
@@ -192,5 +191,169 @@
                 "dataKey": "patients"
             }
         ]
+    },
+    "backends": {
```
patrickskowronekdkfz, line 195: See https://github.com/samply/lens/pull/56#pullrequestreview-2062802966

packages/demo/src/AppCCP.svelte
```diff
@@ -24,16 +31,22 @@
...
 
     const measures = [
-        dktkPatientsMeasure,
-        dktkDiagnosisMeasure,
-        dktkSpecimenMeasure,
-        dktkProceduresMeasure,
-        dktkMedicationStatementsMeasure,
-        dktkHistologyMeasure,
+        {
+            name: "DKTK",
```
patrickskowronekdkfz, line 35: If I understand this correctly is the name associtated with the Measures? In that case it would work to use different measures with a backend. If not then we will have issues, eg. spot needs different measures then spot.

packages/demo/src/AppCCP.svelte
```diff
@@ -46,69 +59,18 @@
...
-        ["mainz", "Mainz"],
-        ["muenchen-lmu", "München(LMU)"],
-        ["muenchen-tum", "München(TUM)"],
-        ["ulm", "Ulm"],
-        ["wuerzburg", "Würzburg"],
-        ["mannheim", "Mannheim"],
-        ["dktk-test", "DKTK-Test"],
-        ["hamburg", "Hamburg"],
-    ];
+    // // VITE_TARGET_ENVIRONMENT should be set by the ci pipeline
```
patrickskowronekdkfz, line 62: Can be deleted?

packages/demo/src/AppCCP.svelte
```diff
@@ -120,6 +82,62 @@
...
+            dktkMedicationStatementsMeasure,
+            dktkHistologyMeasure,
+        ] as MeasureItem[];
+
+        const criteria = dataPasser.getCriteriaAPI("diagnosis");
+
+        const cql = translateAstToCql(
+            ast,
+            false,
+            "DKTK_STRAT_DEF_IN_INITIAL_POPULATION",
```
patrickskowronekdkfz, line 107: Should this not come from the config? Do I need to do this for every sub project? Could we provide a interface?

patrickskowronekdkfz (changes requested):

packages/demo/src/AppCCP.svelte
MatsJohansen87: The Idea is to have multiple spots with each different measures. the name (here DKTK) is to identify which measure belongs to which spot/blaze

packages/demo/src/AppCCP.svelte
MatsJohansen87: this was a showcase for using a custom backend. spots can be configured in the options to work out of the box. this code has been removed

packages/demo/src/AppFragmentDevelopment.svelte
```diff
@@ -149,6 +94,63 @@
...
+
+        console.log("app cql translation", measureItems);
+        const cql = translateAstToCql(
+            ast,
+            false,
+            "DKTK_STRAT_DEF_IN_INITIAL_POPULATION",
+            measureItems,
+            criteria,
+        );
+        console.log("app build meausre", measures);
```
1000daysofcode, line 125: Typo -> "measure"

packages/demo/public/options-ccp-demo.json
```diff
@@ -0,0 +1,305 @@
...
+                        "therapy_of_tumor",
+                        "ProcedureType"
+                    ],
+                    [
+                        "75186-7",
+                        "75186-7"
+                    ]
+                ]
+            }
+        ]
```
torbrenner, line 303: Why are you adding this to the demo project, which is intended to be hosted on demo.lens.samply.de? Wouldn't it make only sense on options-dev.json?

docker-compose.dev.yml
```diff
@@ -21,11 +21,13 @@ services:
   spot:
     image: samply/rustyspot:main
     ports:
-      - 8080:8080
+      - 8055:8055
     environment:
-      BEAM_SECRET: "${LOCAL_BEAM_SECRET_SPOT}"
+      RUST_LOG: "info"
+      CORS_ORIGIN: http://localhost:5173
```
torbrenner, line 27: I thinki this is only necessary in local testing and therefore should be part of your local docker-compose.override.yml.

docker-compose.dev.yml
```diff
@@ -21,11 +21,13 @@ services:
...
+      BEAM_SECRET: "${LOCAL_BEAM_SECRET_FOCUS}"
       BEAM_PROXY_URL: http://beam-proxy:8081
-      BEAM_APP_ID: "spot.${LOCAL_BEAM_ID}.${BROKER_HOST}"
+      BEAM_APP_ID: "focus.${LOCAL_BEAM_ID}.${BROKER_HOST}"
```
torbrenner, line 30: Why are you renaming spot to focus?

torbrenner (changes requested):

packages/demo/public/options-ccp-demo.json
MatsJohansen87: As the backend url is now in the config, it's not possible to have these set by the environment in json, afaik. So we would have to distinguish between imported files rather than adding the string.

so far:
 const backendUrl =
        import.meta.env.VITE_TARGET_ENVIRONMENT === "production"
            ? "https://backend.data.dktk.dkfz.de/prod/"
            : "https://backend.demo.lens.samply.de/prod/";

new change:
  let optionsFilePath: string = "options-dev.json";

    if (import.meta.env.VITE_TARGET_ENVIRONMENT === "production") {
        optionsFilePath = "options-ccp-prod.json";
    } else if (import.meta.env.VITE_TARGET_ENVIRONMENT === "staging") {
        optionsFilePath = "options-ccp-demo.json";
    }

does that make sense, or is there a cleaner way?

docker-compose.dev.yml
MatsJohansen87: Then i'll revert that. Is BEAM_SECRET needed in prod?

docker-compose.dev.yml
MatsJohansen87: i thought we did this together. was that also a local / prod thing? i usually don't temper with the dockerfiles on my own...

torbrenner: i just checked the production configuration, there we still have "spot" as id. So i would change it back. Can you also change back `APP_focus_KEY` to `APP_spot_KEY` in the beam-proxy environment section?

docker-compose.dev.yml
torbrenner: `BEAM_SECRET` is still needed

packages/demo/public/options-ccp-demo.json
torbrenner: Sorry, i think i should have been more specific. The solution with the different config files per project is really good and i like that approach. I thought github would show the actual section i marked for my comment, but that is not the case. I actually wanted to ask why you add a blaze configuration for URL http://localhost:8080 to the production environment?

torbrenner (approved):

PR #96, opened 2024-05-22
fix: remove c% and d% from catalog, which will not be resolved

patrickskowronekdkfz:
### Description
This PR removes all the c% and d% form the catalog
### Motivation and Context

Currently the ast2cql parser does not resolve the c% icd 10 codes. Removing has no impact on the search

### How Has This Been Tested?

locally with spot+beam+sites


Threated (approved):
LGTM

PR #105, opened 2024-06-18
Fix/catalogue scroll behaviour

MatsJohansen87:
### General Summary
Fix the scroll behaviour with sticky elements.

### Motivation and Context
Elements at the bottom of the catalgoue could only be reached when the whole page was scrolled to bottom .
The header is now also sticky while the footer is now at the bottom when scrolling down.


1000daysofcode (approved):
LGTM

PR #127, opened 2024-09-09
fix: use proper options file for production

torbrenner:
Production Lens currently fails to load the options.json file.


torbrenner:
verify new release fails as it compares develop and main (we are skipping develop in this case as it is urgent). I increased the version number to 0.3.1

MatsJohansen87 (approved):

PR #129, opened 2024-09-12
feat(parameters): change attributes to parameters

MatsJohansen87:
### General Summary
There was some confusion when to pass a string and when to pass a javascript object into the components.
Now there will be only JS types (including strings) and there will be no need to JSON.stringify() values passt to the component.

### This is a breaking change!

### DIFF
--- a/packages/demo/src/AppCCP.svelte
+++ b/packages/demo/src/AppCCP.svelte
@@ -13,8 +13,10 @@
         dktkHistologyMeasure,
     } from "./measures";
 
-    let catalogueData: string = "";
-    let libraryOptions: string = "";
+    /**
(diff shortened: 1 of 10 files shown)


1000daysofcode (commented):
Maybe I'm missing something, but why were `props` removed from all of these components (such as with the `InfoButtonComponent`?

Other than that, the changes based on the commit message seem good

MatsJohansen87:
> Maybe I'm missing something, but why were `props` removed from all of these components (such as with the `InfoButtonComponent`?
> 
> Other than that, the changes based on the commit message seem good

The `props` tells the component to use html attributes (which takes only a string). removing this means, you can put in the data as js/ts and don't have to stringify it before.

patrickskowronekdkfz (approved):

PR #154, opened 2024-11-22
Implemented re-import of the Ast

torbrenner:
### General Summary

In the first implementation of lens it was possible to re-import an AST back into lens. With this implementation, the same will be possible in this implementation.
The new implementation is currently in one point inferior to the old implementation. With the old implementation, the main id of the query didn't change upon import. This caveat is currently caused by the ast not containing the query id. This topic should be clarified before the merge.

---

### How Has This Been Tested?
Currently by locally comparing a query before transforming it to ast with after the import of that same query

---

### DIFF
--- a/packages/lib/src/helpers/ast-transformer.ts
+++ b/packages/lib/src/helpers/ast-transformer.ts
@@ -1,5 +1,19 @@
-import type { AstElement, AstTopLayer } from "../types/ast";
-import type { QueryItem, queryStoreItem } from "../types/queryData";
+import { v4 as uuidv4 } from "uuid";
+import {
+    isTopLayer,
+    type AstElement,
+    type AstTopLayer,
+    isBottomLayer,
+    type AstBottomLayerValue,
+} from "../types/ast";
+import type { QueryItem, QueryValue, queryStoreItem } from "../types/queryData";
+import type { AggregatedValue } from "../../../../dist/types";
+import {
+    catalogue,
+    getCategoryFromKey,
+    getCriteriaFromKey,
+} from "../stores/catalogue";
+import { get } from "svelte/store";
 
 /**
  * builds an AST from the query store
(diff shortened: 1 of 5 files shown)


packages/lib/src/types/dataPasser.ts
```diff
@@ -30,4 +30,5 @@ export interface LensDataPasser extends HTMLElement {
     removeValueFromQueryAPI(params: RemoveValueFromQueryAPIParams): void;
     updateResponseAPI(params: ResponseStore): void;
     setQueryStoreAPI(params: QueryItem[][]): void;
+    setQueryStoreFromAstAPI(params: QueryItem[][]): void;
```
patrick-skowronek, line 33: This function name does not match the one in the definition. Literally can't test 😉 The Rest of the code looks good.

packages/lib/src/helpers/ast-transformer.ts
```diff
@@ -19,6 +33,197 @@ export const buildAstFromQuery = (queryStore: QueryItem[][]): AstTopLayer => {
...
+const convertComplexCriteria = (criteria: AstTopLayer): QueryItem => {
+    const values = criteria.children.filter((child) => isTopLayer(child));
+    const checkedValues: AstTopLayer[] = <AstTopLayer[]>values;
+    const convertedValues = convertValues(checkedValues);
+    const criteriaCateogory = getCategoryFromKey(get(catalogue), criteria.key);
+    return {
+        id: uuidv4(),
+        key: criteria.key,
+        name: criteriaCateogory != undefined ? criteriaCateogory.name : "",
+        // NOTE: Assuming hard coding here, proof me wrong
```
patrick-skowronek, line 114: For min/max we have between, see here https://github.com/samply/lens/blob/2b251f0bc14e490122eb19d17cba428b907b2001/packages/demo/public/catalogues/catalogue-dktk.json#L7117. Disproved!

patrick-skowronek: I also think we have somewhere a enum for all the types

patrick-skowronek (changes requested):

packages/lib/src/types/dataPasser.ts
torbrenner: Okay, i will adjust the internal name in the datapasser web component to match the one mentioned here.

packages/lib/src/helpers/ast-transformer.ts
torbrenner: In this case we don't get a BETWEEN condition. The hard coding is only for the "complexCriterias" (multi-condition criterias, like dktk's tumor entities). In this case i think the top part should always be an "EQUALS" condition while all values below are parsed separatly as criteria.

patrick-skowronek (approved):

PR #156, opened 2024-12-10
Ensure catalogue search starts only with 3 inserted character

torbrenner:
This ensures the page doesn't freeze then starting to type in the search bar.

### DIFF
--- a/packages/lib/src/components/search-bar/SearchBarComponent.wc.svelte
+++ b/packages/lib/src/components/search-bar/SearchBarComponent.wc.svelte
@@ -28,6 +28,8 @@
      * @param noMatchesFoundMessage takes a string to display when no matches are found
      */
     export let noMatchesFoundMessage: string = "No matches found";
+    export let typeMoreMessage: string =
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -432,6 +434,10 @@
                 <li>{noMatchesFoundMessage}</li>
             {/if}
         </ul>
+    {:else if autoCompleteOpen && inputValue.length > 0 && inputValue.length < 2}
```
patrick-skowronek: The maximum length should be 3

patrick-skowronek (changes requested):
It is more useable than before. From a user perspective the text disappears when you type 2 letters and searches when typed 3.

packages/lib/src/components/search-bar/SearchBarComponent.wc.svelte
torbrenner: should now be fixed

PR #164, opened 2024-12-18
feat: add message for empty charts

### LINKED ISSUE #160: [Feature]: chart message if no data is available
### General Summary

At the moment there will only be showen a blank chart with no Information when there is no data available to show.

### Suggested Change

Like:
image

### Is this Breaking?

No

### Current Behavior

Empty chart

### New Behavior

Chart with Text [No Data Available]

### Detailed Description

_No response_

valeriesauer:
### Description
Added a "No Data available" message when the chart contains no data. The message is rendered in the chart when all dataset values are 0 or empty

### Related Issue
Closes https://github.com/samply/lens/issues/160#issue-2744494490

### DIFF
--- a/packages/lib/src/components/results/ChartComponent.wc.svelte
+++ b/packages/lib/src/components/results/ChartComponent.wc.svelte
@@ -438,7 +438,7 @@
 
         // If the chart is empty and no responses are pending show "No Data Available"
         noDataAvailable =
-            chartData.labels.length === 0 &&
+            chartData.data[0].data.every((value) => value === 0) &&
             !Array.from(responseStore.values()).some(
                 (response) => response.status === "claimed",
             );
--- a/packages/lib/src/components/results/ChartComponent.wc.svelte
+++ b/packages/lib/src/components/results/ChartComponent.wc.svelte
@@ -79,6 +79,9 @@
     /**
      * initialize the chart
      */
+
+    let noDataAvailable: boolean = false;
+
     let canvas!: HTMLCanvasElement;
 
     let chart: Chart;
@@ -400,7 +403,10 @@
      * @param responseStore - the response store
      */
     const setChartData = (responseStore: ResponseStore): void => {
-        if (responseStore.size === 0) return;
+        if (responseStore.size === 0) {
+            noDataAvailable = true;
+            return;
+        }
 
         let isDataAvailable: boolean = false;
 
@@ -408,7 +414,12 @@
             if (value.status === "succeeded") isDataAvailable = true;
         });
 
-        if (!isDataAvailable) return;
+        if (!isDataAvailable) {
+            noDataAvailable = true;
+            return;
+        }
+
+        noDataAvailable = false;
 
         let chartLabels: string[] = [];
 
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/results/ChartComponent.wc.svelte
```diff
@@ -391,15 +394,23 @@
      * @param responseStore - the response store
      */
     const setChartData = (responseStore: ResponseStore): void => {
-        if (responseStore.size === 0) return;
+        if (responseStore.size === 0) {
```
patrick-skowronek, line 403: The logic is not correct, when the responseStore is empty it should be the "empty charts" look like the initial load. If the Store has results but not regarding this Diagramm then it should show the text

packages/lib/src/components/results/ChartComponent.wc.svelte
```diff
@@ -589,6 +600,11 @@
     {#if options.hintText}
         <InfoButtonComponent message={options.hintText} />
     {/if}
+
+    {#if noDataAvailable}
+        <p>No Data Available</p>
```
patrick-skowronek: Could you make it less aggressive

patrick-skowronek (changes requested):

timakro:
I've changed it such that the "No Data Available" message only shows when no responses are pending. It looks like this now:

image

Notice that the message is missing from the "Therapieart" chart because there is a single data point with the value 0:

```json
{"labels":["medicationStatements"],"data":[{"data":[0],"backgroundColor":["#4dc9f6","#3da4c7"],"backgroundHoverColor":["#aaaaaa"]}]}
```

Instead of checking if "labels" is empty I could alternatively check if all data points are 0. @patrickskowronekdkfz Should I do that?

timakro:
I should be done force pushing now...

patrick-skowronek:
> I've changed it such that the "No Data Available" message only shows when no responses are pending. It looks like this now:
> 
> image
> 
> Notice that the message is missing from the "Therapieart" chart because there is a single data point with the value 0:
> 
> ```json
> {"labels":["medicationStatements"],"data":[{"data":[0],"backgroundColor":["#4dc9f6","#3da4c7"],"backgroundHoverColor":["#aaaaaa"]}]}
> ```
> 
> Instead of checking if "labels" is empty I could alternatively check if all data points are 0. @patrickskowronekdkfz Should I do that?

I think this is the way to go. I'm a bit confused, why we always show a label, even when there is no data

packages/lib/src/styles/result-chart.css
```diff
@@ -1,24 +1,36 @@
...
+  display: grid;
+}
+
+lens-chart::part(chart-wrapper) {
+  display: grid;
+  grid-template-rows: auto 1fr auto;
+  position: relative;
+}
+
+lens-chart::part(chart-overlay) {
```
patrick-skowronek, line 12: For some reason it is not in the center. Could you please check it

patrick-skowronek (changes requested):

packages/lib/src/styles/result-chart.css
timakro: It's at the center of the containing card, which also has the advantage that it's center aligned with the chart title. Centering it on the canvas would make things more difficult because then the canvas would have to be a child of the overlay element (you can't put elements inside a canvas afaik). That in turn would mean that the overlay would always have to be present, not only when the text would show. And that would mean that we have to allow mouse events to go through if text is not showing. Overall I don't think it's worth it. Plus it probably wouldn't look centered in the way you want it either because the charts y axis will make it look lopsided even if it is centered on the canvas.

patrick-skowronek (approved):

Issue #74, opened 2024-03-26
Error Handling in Lens Library


torbrenner:
In the PR #67 we had a short discussion about informing the end user about failed queries etc. As i noted there https://github.com/samply/lens/pull/67#discussion_r1531665422, i think the library part of lens should not directly send error messages to the user, as each application could have their own way of handling such events.
So i think lens should offer a way to send such messages to the end application e.g  by emiting through the [event api](https://developer.mozilla.org/en-US/docs/web/api/event) or alternatively by receiving a communication channel from the application.

timakro:
Do we want user-facing errors when something goes wrong that only a developer can fix?

To provide an example: do we want a user-facing error message when [`options` passed to the `lens-options`](https://github.com/samply/lens/blob/main/docs/SETUP.md#configuration) component is invalid?

patrick-skowronek:
Closed by #171

timakro:
@patrickskowronekdkfz I think we want to show error messages potentially in a few more places (every `console.error()` is a potential candidate). That's why I was asking in which cases.

patrick-skowronek:
You are right, it should be at every `console.error()`. Furthermore, there many lines of code where we don't handle error really well. Also I think it would be nice to have a code snippet to see how it works in the documentation.

timakro:
Are you sure we want it in cases that only a developer can fix? See my previous comment https://github.com/samply/lens/issues/74#issuecomment-2594930207

patrick-skowronek:
> Are you sure we want it in cases that only a developer can fix? See my previous comment [#74 (comment)](https://github.com/samply/lens/issues/74#issuecomment-2594930207)

That is a good question. Since the catalog and options are dynamically loaded there might be problem with the connection. Also there could be an issue with the schema. I, personally, would show it since in development you might now look into the console.

PR #171, opened 2025-01-14
feat: add the option to show user-facing errors as toasts

timakro:
This PR adds the channel `errorChannel` for user-facing error messages. Lens core and applications that use lens core can write to the channel. Applications that use lens core can consume the channel from a custom component or use the provided `<error-toasts />` component to render messages as toasts:

image

Currently only `spot.ts` writes to `errorChannel` - this can be tested by clicking the "Suchen" button in the demo. Before further extending this I wanted to check if you agree with the general approach.

A potential issue is that user-facing messages require localization. For now I've written the error message in `spot.ts` in German, but I suppose not all applications that use lens core are in German.

Related issue: #74

---

### DIFF
--- a/packages/lib/src/classes/spot.ts
+++ b/packages/lib/src/classes/spot.ts
@@ -99,7 +99,7 @@ export class Spot {
             } else {
                 console.error(err);
                 // show user-facing error
-                errorChannel.set(`Fehler beim Bearbeiten der Anfrage: ${err}`);
+                errorChannel.set("Fehler beim Bearbeiten der Anfrage");
             }
         }
     }
--- a/packages/lib/src/classes/spot.ts
+++ b/packages/lib/src/classes/spot.ts
@@ -5,6 +5,7 @@
 import type { SiteData, Status } from "../types/response";
 import type { ResponseStore } from "../types/backend";
 import type { BeamResult } from "../types/spot";
+import { errorChannel } from "../stores/error-channel";
 
 export class Spot {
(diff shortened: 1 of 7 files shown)


packages/lib/src/classes/spot.ts
```diff
@@ -97,6 +98,8 @@ export class Spot {
                 console.log(`Aborting request ${this.currentTask}`);
             } else {
                 console.error(err);
+                // show user-facing error
+                errorChannel.set(`Fehler beim Bearbeiten der Anfrage: ${err}`);
```
patrick-skowronek: I would not report the error to the user. The text is fine.

packages/lib/src/styles/error-toasts.css
```diff
@@ -0,0 +1,17 @@
...
+    right: 0;
+    display: flex;
+    flex-direction: column;
+    gap: var(--gap-xs);
+    padding: var(--gap-s);
+}
+
+error-toasts::part(toast) {
+    border-radius: var(--border-radius-small);
+    background-color: var(--light-orange);
```
patrick-skowronek: I would chose a different color, something more red like #EF9A9A

patrick-skowronek (changes requested):
I think it is missing some small changes. Also could you add a transition of the message disappearing.

patrick-skowronek (approved):
Good work, the next step would be doing it at every console.error and adding a small documentation

PR #173, opened 2025-01-15
feature: added support for logarithmic scale on bar charts

DenisKoether:
added support for logarithmic scale on bar charts

### DIFF
--- a/packages/lib/src/components/results/ChartComponent.wc.svelte
+++ b/packages/lib/src/components/results/ChartComponent.wc.svelte
@@ -39,7 +39,7 @@
     export let headers: Map<string, string> = new Map<string, string>();
     export let displayLegends: boolean = false;
     export let chartType: keyof ChartTypeRegistry = "pie";
-    export let scaleType: "logarithmic";
+    export let scaleType: string = "linear";
     export let perSite: boolean = false;
     export let groupRange: number = 0;
     export let groupingDivider: string = "";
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/results/ChartComponent.wc.svelte
```diff
@@ -39,6 +39,7 @@
     export let headers: Map<string, string> = new Map<string, string>();
     export let displayLegends: boolean = false;
     export let chartType: keyof ChartTypeRegistry = "pie";
+    export let scaleType: "logarithmic";
```
patrick-skowronek: Default should be Linear

packages/lib/src/components/results/ChartComponent.wc.svelte
```diff
@@ -124,6 +125,7 @@
             scales: {
                 y: {
                     display: viewScales,
+                    type: scaleType,
```
patrick-skowronek: In the case we switch the x and y axis's like in the diagnosis on the demo project. This will break the view, since the Y axis has labels instead of values.

patrick-skowronek (changes requested):

packages/lib/src/components/results/ChartComponent.wc.svelte
DenisKoether: Should logarithmic still be possible in the swapped view?

patrick-skowronek: Yes

patrick-skowronek (approved):

PR #177, opened 2025-01-23
Fix lots of typescript errors

timakro:
I want to push for fixing all typescript errors asap and enabling CI such that no new errors can be added. I've fixed a bunch here. I've tried to be careful and not break anything but this nevertheless needs thorough testing. Unfortunately I haven't been able to test with data yet. I'm not too familiar with lens so I would be happy for your input how a thorough test could look.

EDIT: I've addressed all type errors in `packages/lib`. Some errors are still remaining in `packages/demo` and `vite.config.ts`.


timakro:
A few of my fixes are more silencing the type checker and less getting at the root issue. In these cases I left `TODO` comments with suggestions for improvement. I also noticed some places where the current typing is weak. For example the `Site` type in `response.ts` is a common pattern:

```
export type Status = "claimed" | "succeeded" | "tempfailed" | "permfailed";

export type Site = {
    status: Status;
    data?: SiteData;
};
```

This can be success in which case there will be data or failure in which case there won't be. This should rather be a tagged union to encode this into the type system.

packages/lib/src/components/buttons/SearchButtonComponenet.wc.svelte
```diff
@@ -49,13 +49,6 @@
 
         const ast = buildAstFromQuery($queryStore);
 
-        if (ast.children.includes(null)) {
-            alert(
-                "No query entered in one of the queries. You can enter a query or remove the query fields.",
```
patrick-skowronek, line 55: Please use the error channel to inform the User that one search bar is empty

timakro:
@patrickskowronekdkfz Is it ok that I deleted `AppFragementDevelopment.svelte`? Besides that it's ready to merge.

timakro:
Actually, hold off on merging I need to fix something

timakro:
I can't test with data because `options-dev.json` does not conform to the JSON schema. But you can still merge, it should be fine.

patrick-skowronek (changes requested):

packages/demo/src/AppCCP.svelte
```diff
@@ -127,7 +127,7 @@
 <main>
     <div class="search">
         <div class="search-wrapper">
-            <lens-search-bar
+            <lens-search-bar-multiple
```
patrick-skowronek: In AppCCP we don't want to use the single search bar, just one last small thing

patrick-skowronek (approved):

PR #181, opened 2025-01-28
docs: document how to test with data

timakro:
### Related Issue

---

### Screenshots (if appropriate):

---

- [ ] The commit message follows guidelines


patrick-skowronek (changes requested):
I have some thoughts about this PR. I would, if possible, not include any binary data. Also with the moving of the CCP-Explorer in the next version we should have documentation who to use the demo project with a Blaze, Spot or just sending AST. This is a good start, but maybe we should discuss what the documentation should include

timakro:
Yeah I agree that this is far from what we should aim for in terms of testing convenience. Honestly I wrote this mostly for myself to remember the steps. We need something better in the long term.

PR #182, opened 2025-01-30
feat: added info link for catalogue catagories

patrick-skowronek:
### Description
We need the possibility to add link to certain catalogue entries. This PR added infoLink as new add option and generated the link beside the info button.

### DIFF
--- a/packages/lib/src/types/treeData.ts
+++ b/packages/lib/src/types/treeData.ts
@@ -11,6 +11,12 @@ export type CategoryNode = {
     childCategories?: Category[];
     infoButtonText?: string[];
     subCategoryName?: string;
+    extra?: CategoryExtra;
+};
+
+export type CategoryExtra = {
+    link?: string;
+    display?: string;
 };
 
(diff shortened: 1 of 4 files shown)


packages/lib/src/types/treeData.ts
```diff
@@ -11,6 +11,12 @@ export type CategoryNode = {
     childCategories?: Category[];
     infoButtonText?: string[];
     subCategoryName?: string;
+    extra?: CategoryExtra;
```
timakro: Shouldn't this be `infoLink` instead of `extra`?

packages/lib/src/types/treeData.ts
```diff
@@ -11,6 +11,12 @@ export type CategoryNode = {
...
+};
+
+export type CategoryExtra = {
+    link?: string;
+    display?: string;
```
timakro: These are required in the JSON schema so why not make them required here as well?

timakro (approved):

PR #186, opened 2025-02-03
chore: turn Status into a tagged union

timakro:


packages/lib/src/stores/response.ts
```diff
@@ -73,66 +68,41 @@ export const getSitePopulationForCode = (
...
-    const population: number[] = [];
-    if (store.size === 0) return 1;
-
-    sites.forEach((site: Site) => {
-        population.push(
-            getSitePopulationForStratumCode(
-                site.data!,
+    if (store.size === 0) {
+        return 1;
+    }
```
timakro: Does anyone know why we have this special case? I want to add a comment that explains this

timakro:
Please merge #204 first. There might be conflicts and #204 should take priority.

patrick-skowronek (approved):

PR #190, opened 2025-02-03
fix: indicate sorting in table headers using arrows

### LINKED ISSUE #185: [Bug]: Result table sorting partially implemented
### General Summary

Result table can be sorted. This shall be made visible to the user (i.e. currently no links are shown for table headers).

Each column can be only sorted ascending at the moment.

### Expected Behavior

If sorted for one column, the column header shall be highlighted in a way showing this and the mouse cursor shall change when moving over a column header.

All columns shall be sortable ascending and descending.

### Current Behavior

No indication the table is sortable. No way to see by which column the table is sorted. No way to sort descending. 

### Possible Solution

_No response_

### Steps to Reproduce

Triger a search. Click on table headers. 

### Context (Environment)

BBMRI-ERIC, DKTK, all environments

### Detailed Description

_No response_

timakro:
Fixes #185

### DIFF
--- a/packages/lib/src/components/results/ResultTableComponent.wc.svelte
+++ b/packages/lib/src/components/results/ResultTableComponent.wc.svelte
@@ -145,6 +145,13 @@
 
         // sort
         tableRowsCopy.sort((a, b) => {
+            // Always sort claimedText below everything else
+            if (a[sortColumnIndex] === claimedText) {
+                return 1;
+            } else if (b[sortColumnIndex] === claimedText) {
+                return -1;
+            }
+
             if (a[sortColumnIndex] < b[sortColumnIndex]) {
                 return sortAscending ? -1 : 1;
             }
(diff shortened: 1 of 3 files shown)


timakro:
image

packages/lib/src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -213,45 +190,42 @@
...
-                        sortTable(index, header.ascending, tableRowData, true)}
+                    on:click={() => clickedOnColumnHeader(index)}
                 >
                     {header.title}
                     {#if header.hintText}
                         <InfoButtonComponent message={header.hintText} />
                     {/if}
+                    {#if index === sortColumnIndex}
+                        {#if sortAscending}
+                            ▲
```
patrick-skowronek: Please make the font size a bit smaller

packages/lib/src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -155,47 +133,46 @@
...
+    $: {
+        // Array.sort sorts in place, so make a copy first
+        const tableRowsCopy = [...tableRowData];
 
-        tableRowData = tableRowData.sort((a, b) => {
-            if (a[column] < b[column]) {
-                return ascending ? -1 : 1;
+        // sort
+        tableRowsCopy.sort((a, b) => {
+            if (a[sortColumnIndex] < b[sortColumnIndex]) {
```
patrick-skowronek, line 155: While some sites are still processing, they don't get sorted. If you sort by a number, eg. patient count, the processing should always be at the end

patrick-skowronek (changes requested):
I like the that now both ways to sort works, some small issues are there still

patrick-skowronek (approved):

PR #192, opened 2025-02-04
chore: remove problematic CQL line from patientsMeasure

timakro:
I think we wanted to remove this CQL line @patrickskowronekdkfz?


patrick-skowronek (approved):

patrick-skowronek:
Yes, this line does not work properly and is confusing for users

PR #214, opened 2025-02-11
feat: also replace % in aggregatedValues

patrick-skowronek:
### General Summary

This PR replace subgroup placeholders in aggregateValues. This is needed for the catalogue of CCP-Explorer and future Projects

### DIFF
--- a/packages/lib/src/cql-translator-service/ast-to-cql-translator.ts
+++ b/packages/lib/src/cql-translator-service/ast-to-cql-translator.ts
@@ -13,11 +13,10 @@ import {
     cqltemplate,
     criterionMap,
 } from "./cqlquery-mappings";
-import { getCriteria, resolveAstSubCatagories } from "../stores/catalogue";
+import { resolveAstSubCatagories } from "../stores/catalogue";
 import type { MeasureItem } from "../types/backend";
 
-let codesystems: string[] = [];
-let criteria: string[] = [];
+const codesystems: string[] = ["codesystem loinc: 'http://loinc.org'"];
 
 export const translateAstToCql = (
     query: AstTopLayer,
@@ -25,11 +24,6 @@ export const translateAstToCql = (
     backendMeasures: string,
     measures: MeasureItem[],
 ): string => {
-    if (criteria.length == 0) {
-        criteria = getCriteria("diagnosis");
-        codesystems = ["codesystem loinc: 'http://loinc.org'"];
-    }
-
     const localMeasures: { key: string; cql: string }[] = [];
     measures.forEach((x) => {
         localMeasures.push({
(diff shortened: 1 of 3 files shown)


timakro (approved):

timakro:
Very nice change, now the % replacement is all in one place.

This means we can also get rid of this call:
https://github.com/samply/lens/blob/42fe9310f2147554ca4b94b7473bebbc92ca4dd9/packages/lib/src/cql-translator-service/ast-to-cql-translator.ts#L29

Maybe we could also completely get rid of `getCritera()` and its helper functions? That would only work though insofar nobody is using this interface:
https://github.com/samply/lens/blob/42fe9310f2147554ca4b94b7473bebbc92ca4dd9/packages/lib/src/components/DataPasser.wc.svelte#L69-L71

patrick-skowronek:
> Very nice change, now the % replacement is all in one place.
> 
> This means we can also get rid of this call:
> 
> https://github.com/samply/lens/blob/42fe9310f2147554ca4b94b7473bebbc92ca4dd9/packages/lib/src/cql-translator-service/ast-to-cql-translator.ts#L29
> 

You're right. This is then no longer needed and was only for a workaround. I will remove it.

> Maybe we could also completely get rid of `getCritera()` and its helper functions? That would only work though insofar nobody is using this interface:
> 
> https://github.com/samply/lens/blob/42fe9310f2147554ca4b94b7473bebbc92ca4dd9/packages/lib/src/components/DataPasser.wc.svelte#L69-L71

I would not remove it yet. This is no longer used in the core, however it might be used in a project. Therefore I would maybe mark it as deprecated.

packages/lib/src/cql-translator-service/ast-to-cql-translator.ts
```diff
@@ -13,23 +13,17 @@ import {
     cqltemplate,
     criterionMap,
 } from "./cqlquery-mappings";
-import { getCriteria, resolveAstSubCatagories } from "../stores/catalogue";
+import { resolveAstSubCatagories } from "../stores/catalogue";
 import type { MeasureItem } from "../types/backend";
 
-let codesystems: string[] = [];
```
timakro, line 19: The `codesystems` variable is still used elsewhere, no?

timakro (changes requested):

timakro:
Marking `getCriteriaAPI` as deprecated sounds good to me.

packages/lib/src/cql-translator-service/ast-to-cql-translator.ts
patrick-skowronek: Yes it is in line 504, 505 and so on. It is a global variable and not reset after the function is called. However, having CodeSystem which are not used is not crashing anything. Your are right, the code style there is bad. However, we should improve it in the future.

timakro: Ah my bad, I misread the diff. Looks good.

timakro (approved):

Issue #207, opened 2025-02-11
[Bug]: Negotiator Service generates wrong json, when all collections are selected


patrick-skowronek:
### General Summary

We observed that after selecting each site for negotiation the output of the JSON.stringify is not valid JSON.

### Expected Behavior

Output correct JSON

### Current Behavior

This is what is printed out `{
  "humanReadable": "",
  "url": "http://localhost:5173",
  "resources":
    [
      {
        "id": "bbmri-eric:ID:DE_RWTHCBMB:collection:RWTHCBMB_BC",
        "name": "bbmri-eric:ID:DE_RWTHCBMB:collection:RWTHCBMB_BC",
        "organization":
          {
            "id": 0,
            "externalId": "bbmri-eric:ID:DE_RWTHCBMB",
            "name": "Aachen",
          },
      },
      {
        "id": "bbmri-eric:ID:DE_ACBB:collection:LIQUID",
        "name": "bbmri-eric:ID:DE_ACBB:collection:LIQUID",
        "organization":
          {
            "id": 0,
            "externalId": "bbmri-eric:ID:DE_ACBB",
            "name": "Augsburg",
          },
      },
      {
        "id": "bbmri-eric:ID:DE_ZeBanC:collection:COVID19",
        "name": "bbmri-eric:ID:DE_ZeBanC:collection:COVID19",
        "organization":
          {
            "id": 0,
            "externalId": "bbmri-eric:ID:DE_ZeBanC",
            "name": "Berlin",
          },
      },
      {
        "id": "bbmri-eric:ID:CZ_MMCI:collection:DNA",
        "name": "bbmri-eric:ID:CZ_MMCI:collection:DNA",
        "organization":
          { "id": 0, "externalId": "bbmri-eric:ID:CZ_MMCI", "name": "Brno" },
      },
      {
        "id": "bbmri-eric:ID:CY_cybiobank:collection:General_Biobanking_cohort",
        "name": "bbmri-eric:ID:CY_cybiobank:collection:General_Biobanking_cohort",
        "organization":
          {
            "id": 0,
            "externalId": "bbmri-eric:ID:CY_cybiobank",
            "name": "Cyprus",
          },
      },
      {
        "id": "bbmri-eric:ID:DE_BBD:collection:DILB",
        "name": "bbmri-eric:ID:DE_BBD:collection:DILB",
        "organization":
          { "id": 0, "externalId": "bbmri-eric:ID:DE_BBD", "name": "Dresden" },
      },
      {
[...]

PR #215, opened 2025-02-11
feat: improve BBMRI negotiator error handling and do some cleanup

### LINKED ISSUE #210: [Feature]: Use error channel and cover all http statues for negotiatorservice
### General Summary

Currently there are more http codes then we handle, see https://negotiator.bbmri-eric.eu/api/swagger-ui/index.html#/Requests/add . Also, we should not use an alert when error occurs.

### Suggested Change

Add error channel

### Is this Breaking?

No

### Current Behavior

Alert

### New Behavior

- 

### Detailed Description

_No response_

timakro:
Important: I couldn't fully test this because I'm not whitelisted to send requests to the BBMRI Negotiator.

I set out to improve the error handling and ended up rewriting the entire file. Improvements include:
* HTTP status codes other than 201 are now always logged to the developer console including the response body. This would have saved us some debugging when working on https://github.com/samply/lens/issues/207.
* Use `errorChannel` instead of `alert()` to inform the user of errors
* Simplified how lens options are accessed
* Fixed the name of `negotiateOptions` (previously incorrectly `negotiatorOptions`) so that types are correctly inferred
* (Hopefully) improved the documentation

I've decided to use the same user-facing error message for all types of errors: "Fehler beim Anfragen der Daten und Proben". My take is that describing in more detail what went wrong (e.g. network error or unexpected HTTP status code) wouldn't really provide any benefit to the user who couldn't take action anyways. So this potentially fixes https://github.com/samply/lens/issues/210 even though it doesn't do exactly what is described there because I don't think it's a good idea. Do you agree in general with this approach or do you prefer more specific user-facing errors?

### DIFF
--- a/packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
+++ b/packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
@@ -16,9 +16,9 @@
      *
      */
     function exec(): void {
-        if (type == "bbmri") {
+        if (type == "Negotiator") {
             bbmriNegotiate($datarequestsStore);
-        } else if (type == "ccp") {
+        } else if (type == "ProjectManager") {
             negotiate($datarequestsStore);
         }
     }
--- a/packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
+++ b/packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
@@ -7,7 +7,7 @@
 <script lang="ts">
     import { datarequestsStore } from "../../stores/datarequests";
     import { negotiate } from "../../services/ccpProjectManager";
-    import { bbmrinegotiate } from "../../services/bbmriNegotiate";
+    import { bbmriNegotiate } from "../../services/bbmriNegotiate";
 
     export let title: string = "Negotiate with biobanks";
     export let type: string;
@@ -17,7 +17,7 @@
      */
     function exec(): void {
         if (type == "bbmri") {
-            bbmrinegotiate($datarequestsStore);
+            bbmriNegotiate($datarequestsStore);
         } else if (type == "ccp") {
             negotiate($datarequestsStore);
         }
--- a/packages/lib/src/services/bbmriNegotiate.ts
+++ b/packages/lib/src/services/bbmriNegotiate.ts
@@ -55,9 +55,16 @@ export async function bbmriNegotiate(
     }
 
     const bbmriCollectionResources: BbmriCollectionResource[] = [];
-    for (const siteMapping of currentLensOptions.negotiateOptions
-        .siteMappings) {
-        if (sitesToNegotiate.includes(siteMapping.site)) {
+    for (const site of sitesToNegotiate) {
+        const siteMapping =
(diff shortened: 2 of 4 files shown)


packages/lib/src/services/bbmriNegotiate.ts
```diff
@@ -1,164 +1,127 @@
...
     try {
-        response = await fetch(`${negotiateOptions.url}`, {
+        // Swagger documentation: https://negotiator.bbmri-eric.eu/api/swagger-ui/index.html#/Requests/add
+        response = await fetch(url, {
             method: "POST",
             headers: {
                 Accept: "application/json; charset=utf-8",
                 "Content-Type": "application/json",
                 Authorization:
                     "Basic YmJtcmktZGlyZWN0b3J5Omw5RFJLVUROcTBTbDAySXhaUGQ2",
```
patrick-skowronek: While on a cleanup, this should also be set via the options. I know that it will be removed at some time.

packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
```diff
@@ -17,7 +17,7 @@
      */
     function exec(): void {
         if (type == "bbmri") {
```
patrick-skowronek: When we break the options anyway. We should also rename this to "NEGOTIATOR" and "PROJECTMANAGER". Please add https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change to your commit

patrick-skowronek (changes requested):

packages/lib/src/services/bbmriNegotiate.ts
timakro: What do you mean? The swagger URL? The `negotiateOptions.url` is still configurable in the lens options, it's just passed to that function as `url`.

packages/lib/src/components/buttons/NegotiateButtonComponent.wc.svelte
timakro: Honestly I'd feel more comfortable doing this when we have the typescript and JSON schema conversion in place and are running the options checker/catalogue checker again.

I'll start two discussion threads where we can collect breaking changes we want to make (one for options and one for catalogue).

timakro: A suggestion: We could use PascalCase for enum-like strings everywhere. This would work well for tagged union discriminators because we can use the name of the typescript type (which is in PascalCase). And Rusts serde likes this as well when parsing enums.

packages/lib/src/services/bbmriNegotiate.ts
patrick-skowronek: I pined the comment on line 149/107, there the basic auth password is shown. In the future I would prefer to set it to empty instead of setting it directly in the code

timakro:
Should we show an error toast if a site that has its checkbox checked in the table cannot be found in `negotiateOptions.siteMappings` in the lens options? If not, should we show an error in the developer console?

patrick-skowronek (approved):
Looks good, we should start documenting the API's

PR #217, opened 2025-02-12
chore: deprecate getCriteriaAPI

### LINKED ISSUE #216: [Feature]: mark getCriteria as deprecated
### General Summary

This function is no longer needed in the core, there might be a usage outside of the core library. We should mark it as deprecated and wait until a major version release

### Suggested Change

Mark it as deprecated

### Is this Breaking?

Yes

### Current Behavior

-

### New Behavior

-

### Detailed Description

_No response_

### LINKED ISSUE #222: [Feature]: Add Library function to get catalogue
### General Summary

The function getCriteriaAPI returns a part of the catalogue, with #217 deprecated and should be remove in v0.4.7. However, this function should not be lost. 

I would rather get a subscribable writeable of the catalogue to modify or search the catalogue. 

### Suggested Change

Add getCatalog()

### Is this Breaking?

Yes

### Current Behavior

getCriteria(String)

### New Behavior

-

### Detailed Description

-

timakro:
Closes #216
Closes #222

### DIFF
--- a/packages/lib/src/components/DataPasser.wc.svelte
+++ b/packages/lib/src/components/DataPasser.wc.svelte
@@ -65,6 +65,7 @@
      * returns the catalogue to the library user
      * @param category the category name (e.g. "diagnosis")
      * @returns array of strings containing the bottom level items' keys
+     * @deprecated may be removed in a future version
      */
     export const getCriteriaAPI = (category: string): string[] => {
         return getCriteria(category);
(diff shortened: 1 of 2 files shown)


patrick-skowronek (changes requested):
This looks good, please add that it is marked to be deleted in version v0.4.7 and you should use #222 instead.

patrick-skowronek (approved):

PR #231, opened 2025-02-26
Automatically generate catalogue JSON schema

timakro:
This adds automatic JSON schema generation for the catalogue. The lens options are hopefully to follow at some point.

Heads up for reviewer: it is probably easier to review this PR commit by commit.

I've given the catalogue type definitions the same treatment as we are planning for the AST: made it a proper tagged union and split it into more fine-grained variants. I also added a bunch of JSDoc to the type definitions.

There is one breaking change that I have marked in conventional commit convention. The catalogue JSON now needs an additional `"fieldType": "group"` on all group objects, i.e. those that have `"childCategories"`.

You can regenerate the JSON schema with `npm run schemagen`.

CI checks that

1. the JSON schema matches the typescript definitions and
2. the JSON files in the project match the JSON schema.


packages/lib/src/helpers/ast-transformer.ts
```diff
@@ -99,9 +99,6 @@ const convertCriteria = (criteria: AstTopLayer): QueryItem | undefined => {
         system: checkedValues[0].system,
         values: convertedValues,
     };
-    if (catalogueCategory != undefined && "description" in catalogueCategory) {
-        convertedCriteria.description = catalogueCategory.description;
-    }
```
timakro, line 104: Note to reviewer: removed this because Category never has the description property

packages/lib/src/helpers/ast-transformer.ts
```diff
@@ -121,10 +118,6 @@ const convertComplexCriteria = (criteria: AstTopLayer): QueryItem => {
         name: criteriaCategory != undefined ? criteriaCategory.name : "",
         // NOTE: Assuming hard coding here, proof me wrong
         type: "EQUALS",
-        description:
-            criteriaCategory != undefined && "description" in criteriaCategory
-                ? criteriaCategory.description
-                : "",
```
timakro, line 127: Same as earlier

packages/lib/src/stores/catalogue.ts
```diff
@@ -202,15 +202,6 @@ const itemIsBottomLevel = (item: TreeNode): boolean => {
  * @returns an array of strings containing the bottom level items' keys
  */
 const getBottomLevelItems = (item: TreeNode, category: string): string[] => {
-    /**
-     * FIX ME:
-     *  there seems to be a race condition where the catalogue is not yet loaded and the function is called right away
-     *  the data being a string comes from the data being passed as a json string
-     */
-    if (typeof item === "string") {
```
patrick-skowronek, line 210: This needs to be test in depth. We had issues that item is an empty string and breaks lens. I will investigate.

packages/lib/src/components/catalogue/Catalogue.wc.svelte
```diff
@@ -9,11 +9,13 @@
     import { catalogue } from "../../stores/catalogue";
     import type { ToggleAttribute } from "../../types/helpers";
     import type { CatalogueText } from "../../types/texts";
-    import type { Category } from "../../types/treeData";
+    import type { Catalogue } from "../../types/catalogue";
     import DataTreeElement from "./DataTreeElement.svelte";
     import { iconStore } from "../../stores/icons";
 
-    export let treeData: Category[] = [];
+    // TODO: check if anyone actually uses this, otherwise remove it
```
patrick-skowronek, line 16: Similar to the last comment, I will investigate

packages/lib/src/helpers/ast-transformer.ts
patrickskowronekdkfz: Good point, I think this is a really old data element. I think this was something project specific. Can be removed.

patrick-skowronek: We should think about if inherence of description is a good idea. I would say it is not. I see that we want to efficient in that case. But I would argue it will break more than it will benefit. @timakro This would also be an breaking thing.

packages/lib/src/helpers/ast-transformer.ts
```diff
@@ -121,10 +118,6 @@ const convertComplexCriteria = (criteria: AstTopLayer): QueryItem => {
         name: criteriaCategory != undefined ? criteriaCategory.name : "",
         // NOTE: Assuming hard coding here, proof me wrong
         type: "EQUALS",
-        description:
-            criteriaCategory != undefined && "description" in criteriaCategory
```
patrick-skowronek, line 125: Same here

packages/demo/public/catalogues/catalogue-dktk-staging.json
```diff
@@ -1,5 +1,6 @@
 [
     {
+        "fieldType": "group",
```
patrick-skowronek, line 3: Similar to the change of the AST, do we have much improvement to add this breaking change? How would we migrate to it? Could we automate it, like with a script?

patrick-skowronek (approved):

PR #243, opened 2025-03-10
Svelte5

timakro:
This PR upgrades lens to Svelte 5. I tested BBMRI sample locator and of course CPP explorer and both work.

There are still a bunch of uses of `run()` in the code base which we should replace over time: https://svelte.dev/docs/svelte/v5-migration-guide#Migration-script-run


packages/lib/src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -110,15 +110,19 @@
         });
     };
 
-    $: buildTableRowData($responseStore);
+    run(() => {
```
patrick-skowronek: I believe that this line breaks the new svelte api, see `Detected a migrated `$:` reactive block in `packages/lib/src/components/results/ResultTableComponent.wc.svelte` that both accesses and updates the same reactive value. This may cause recursive updates when converted to an `$effect`. `

packages/lib/src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -110,15 +110,19 @@
...
 
     /**
      * watches the datarequestsStore for changes to check or uncheck the checkbox
      */
-    let allChecked: boolean = false;
-    $: allChecked =
-        $datarequestsStore.length === tableRowData.length &&
-        tableRowData.length !== 0;
+    let allChecked: boolean = $state(false);
+    run(() => {
```
patrick-skowronek: Also this breaks, I can't not select any site or select all, I think this is due to the state been overwritten?

patrick-skowronek (changes requested):

packages/lib/src/components/results/ResultTableComponent.wc.svelte
timakro: Should be fixed

packages/lib/src/components/results/ResultTableComponent.wc.svelte
timakro: Should be fixed

timakro:
Maybe you can test the negotiator/PM, I don't have an option to test it I think

patrick-skowronek (approved):

patrick-skowronek:
> Maybe you can test the negotiator/PM, I don't have an option to test it I think

Good point, I will test it after the merge and will create an issue if errors arise!

PR #253, opened 2025-03-12
Book and typedoc

### LINKED ISSUE #195: [Feature]: Update Svelte to version 5 and Vite to version 6
### General Summary

Svelte 5 is looking pretty good and Vite 6 makes the build better. 

### Suggested Change

Update package.json

### Is this Breaking?

Yes

### Current Behavior

-

### New Behavior

 -

### Detailed Description

_No response_

### LINKED ISSUE #254: [Bug]: Removing criteria doesn't work since Svelte 5
### General Summary

To reproduce:
* Start with an empty search bar
* Add gender=male
* Add gender=female
* Try to remove gender=female (it will not work)

### Expected Behavior

x

### Current Behavior

x

### Possible Solution

x

### Steps to Reproduce

x

### Context (Environment)

x

### Detailed Description

x

timakro:
This adds an mdbook under `book/` and prepares the repo for generating typedoc (https://typedoc.org/) HTML. ~~I haven't added CI yet to build and host it so it's not very useful yet but I will do that soon(tm).~~ I'm thinking GitHub Pages for hosting.

This also fixes another bug introduced by the Svelte 5 update ~~I noticed and there is at least one bug left - I will open an issue for that or fix it tomorrow.~~ I think we should include this PR and also move CCP explorer to its own repo for the next release and make the next release lens 0.5.

This also upgrades all of our dependencies to the latest version.

One major change I've made is that types are now explicitly re-exported in `packages/lib/index.ts` like everything else. This has the advantage that we can control what types are public, it also saves us a bunch of custom code in `vite.config.ts` (that file is very small now) and it means that typedoc will know which types to include in the public facing docs.

As the configuration format of ESLint changed and the old format is deprecated I started from scratch with the ESLint config. It now has only the recommended defaults from TS/JS included in ESLint itself and the recommended Svelte specific lints from `eslint-plugin-svelte`. We could consider adding back the JSDoc lints, although I personally think they weren't really useful.

I also took the time to go through various other configs (e.g. tsconfig.json) and try to figure out the recommended options from the Svelte docs as things may have changed since the repository was created. I've tried to document as good as possible why we have the options we have and added links to upstream docs so we can comprehend in the future why the settings are there and update them if upstream changes their docs.

Closes #195

Closes #254


timakro:
What I forgot to mention: I removed the CommonJS bundle - I think it's safe to assume that in the year 2025 consumers support ESM.

Also the import path of the CSS bundle changed:

```diff
- @import '../node_modules/@samply/lens/dist/style.css';
+ @import '@samply/lens/style.css';
```

timakro:
I just decided to push onto this PR - sorry for making it gigantic...

In case you are wondering how to try out the new `index.html`: run `npx vite`. Vite looks for an `index.html` in the project root by default: https://vite.dev/guide/build#library-mode

index.ts
```diff
@@ -0,0 +1,204 @@
+// This import loads lens CSS and web components as a side effect
+import { setCatalogue, setOptions } from "./packages/lib/src/index";
+import type { QueryEvent, Site } from "./packages/lib/src/index";
+
+setOptions({
+    siteMappings: {
+        bonn: "Bonn",
```
patrick-skowronek: I would use Sites that are not real.

patrick-skowronek (changes requested):

patrick-skowronek:
also closes https://github.com/samply/lens/issues/257

patrick-skowronek (approved):

PR #261, opened 2025-03-14
fix: added default values to date replacements

patrick-skowronek:
### Description

This PR fixes the default values for less than AST2CQL generation


timakro:
Shouldn't we use the `lowerThanDateTemplate` and `greaterThanDateTemplate` if one of min or max is null?

timakro:
Also please remove the `as { min: number; max: number }` cast and just pass in `min: number, max: number` into the function if that is the only thing it accepts.

Apparently it is also sometimes null? Encode this into the type then.

PR #272, opened 2025-03-21
Pr remove ccp explorer

timakro:
Changes:
* remove CCP Explorer
* move Lens code from `packages/lib/src` to `src`
* update README.md, I hope it is okay I deleted so much:
  * I added a link to the Lens Book and API docs
  * The old "Building the library locally" section is replaced by https://github.com/samply/lens-test
  * "Additional Note for Applications using Vite" is no longer needed
  * "Style Integration" should eventually be added to the Lens Book
  * "Roadmap" is pretty much done
  * I moved the License to a separate LICENSE.txt file so GitHub will also detect it and mark it in our repo
* enabled prettier and eslint checks in CI
* formatted everything with prettier


.github/workflows/docker.yml
```diff
@@ -1,37 +0,0 @@
-name: Build and Push Image samply/lens
```
patrick-skowronek, line 1: For automatic testing, I would prefer to run docker in the ci. I would just run the docker build command and try to serve the demo. This is would replace testing, but would show if there is an error while building

docker-compose.yml
```diff
@@ -1,138 +1,138 @@
-version: '3.8'
+version: "3.8"
```
patrick-skowronek: Version can be removed

docker-compose.dev.yml
```diff
@@ -1,76 +1,76 @@
 version: "3.8"
```
patrick-skowronek: Version can be removed

packages/demo/public/catalogues/catalogue-example.json
```diff
@@ -1,128 +0,0 @@
-[
```
patrick-skowronek, line 1: I would keep a catalogue with examples for the small demo

patrick-skowronek (changes requested):

patrick-skowronek (approved):

PR #279, opened 2025-03-24
Feat/improve chart combine subgroup performance

timakro:
`combineSubGroups` in `ChartComponent.wc.svelte` is the reason why BBMRI Sample Locator is unresponsive while search results are coming in. This PR improves the performance a little bit which roughly halves the time the UI is blocked. Although the real fix would be to change the `SiteData` type to allow efficient lookups of stratifier and stratum without searching through everything.

I've also added a source map to the lens package which makes debugging these kind of issues easier with the Chrome Dev Tools.

### DIFF
--- a/src/components/results/ChartComponent.wc.svelte
+++ b/src/components/results/ChartComponent.wc.svelte
@@ -356,68 +356,47 @@
         responseStore: ResponseStore,
         labels: string[],
     ): { labels: string[]; data: number[] } => {
-        const groupedChartData: { label: string; value: number }[] =
-            labels.reduce<{ label: string; value: number }[]>((acc, label) => {
-                // This is a hack! This will help with the wrong coding of ICD10
-                label = label.replace(/_/g, ".");
+        // console.log(`processing ${labels.length} labels`);
+        // const start = performance.now();
 
-                /**
+        const labelsToData = new Map<string, number>();
+        for (const label of labels) {
+            // This is a hack! This will help with the wrong coding of ICD10
+            // label = label.replace(/_/g, ".");
+
+            const value = getAggregatedPopulationForStratumCode(
+                responseStore,
+                label,
+                responseGroupCode,
+            );
+
+            if (!label.includes(divider) || divider === "") {
+                /*
                  * see if the label contains the divider
-                 * if not, add it to the accumulator with a .% at the end
+                 * if not, add it to the accumulator with a grouping label (.%) at the end
                  */
-                if (!label.includes(divider) || divider === "") {
-                    return [
-                        ...acc,
-                        {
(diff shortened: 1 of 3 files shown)


src/components/results/ChartComponent.wc.svelte
```diff
@@ -356,68 +356,47 @@
         responseStore: ResponseStore,
         labels: string[],
     ): { labels: string[]; data: number[] } => {
-        const groupedChartData: { label: string; value: number }[] =
-            labels.reduce<{ label: string; value: number }[]>((acc, label) => {
-                // This is a hack! This will help with the wrong coding of ICD10
-                label = label.replace(/_/g, ".");
+        // console.log(`processing ${labels.length} labels`);
```
patrick-skowronek: Please remove the console.log. I know it is for debugging, but I think it should be in the PR.

src/components/results/ChartComponent.wc.svelte
```diff
@@ -356,68 +356,47 @@
         responseStore: ResponseStore,
         labels: string[],
     ): { labels: string[]; data: number[] } => {
-        const groupedChartData: { label: string; value: number }[] =
-            labels.reduce<{ label: string; value: number }[]>((acc, label) => {
-                // This is a hack! This will help with the wrong coding of ICD10
-                label = label.replace(/_/g, ".");
```
patrick-skowronek, line 362: You are right, this regex replace should not be there. We only have one site left, which reports icd_10 codes with underscores. We can safely delete it.

vite.config.ts
```diff
@@ -9,6 +9,8 @@ export default defineConfig({
             // Only emit ESM, no CommonJS
             formats: ["es"],
         },
+        // Also emit source map as lens.js.map
+        sourcemap: true,
```
patrick-skowronek, line 13: It is really nice for debugging. I'm not quite sure if it will bloat up the npm build. The Maps are quite large with 2mb.

patrick-skowronek (commented):
I tested it locally and it works. I have two concerns. Hope you can address them.

vite.config.ts
timakro: Sure it will make the npm package bigger but I don't see why it would be a problem

patrick-skowronek: If possible, I would keep the system as fast as possible. Adding a big payload to the user would be something I would like to avoid in a production build. But I'm not sure if it will be shipped, when we use Vite. I think we should test it in a beta build.

patrick-skowronek (approved):

vite.config.ts
timakro: The browser will not load the source map if you don't open dev console.

PR #282, opened 2025-03-24
docs: document how to add linting CI and how to configure Prettier

timakro:
Add some docs so that new applications will hopefully do type checking, EsLint and prettier in CI just like Lens.

I've also added a recommendation about using the Prettier defaults. I've changed the indent from tabs to 2 spaces in ccp-explorer which is Prettier's default and the de-facto standard in TS/JS. I'd suggest to do the same in Lens now that moving packages/lib/src/ to src/ broke the git history anyways it would be a good time to do this. Should we do that?


book/src/SUMMARY.md
```diff
@@ -1,9 +1,8 @@
 # Summary
 
-- Components
```
patrick-skowronek, line 3: I think a Components Section would be better. We could add some small documentation about the other components like Catalogue, Backends etc. as well

patrick-skowronek (changes requested):

book/src/SUMMARY.md
timakro: I imagined a split between "User guide" for the users of lens and "Development" for the developers of lens. But yeah it might be sense to make a components section under "user guide"? Or change the structure entirely, I'm not too set on anything. I mostly made this change because SUMMARY.md was just broken before and not building the book. We should probably add book building to CI so we see if the book still builds on every PR.

patrick-skowronek (approved):

patrick-skowronek:
Yeah, we should move to 2 spaces

PR #283, opened 2025-03-24
chore: use uuidv4() instead of crypto.randomUUID()

timakro:
crypto.randomUUID only works in secure contexts which can be an annoyance when testing


src/classes/spot.ts
```diff
@@ -27,7 +28,7 @@ export class Spot {
         controller: AbortController,
     ): Promise<void> {
         try {
-            this.currentTask = crypto.randomUUID();
+            this.currentTask = uuidv4();
```
patrick-skowronek, line 31: Good catch, this might not work certain cases. However, it might be slower. I'm fine with switching

patrick-skowronek (approved):

PR #291, opened 2025-03-25
chore: remove docker compose file and example.env

timakro:
I've moved them to the ccp-explorer repo


patrick-skowronek (approved):
Good cleanup, I think we should add a link to our other components in the docs

PR #293, opened 2025-03-26
feat: added stepSize

### LINKED ISSUE #310: [Bug]: Bar charts should always have a step between the highest value the limit
### General Summary

Currently, there is no intermediate step between the highest value and the diagram’s upper limit on the scale. This can make the scale feel unbalanced.

### Current Behavior

 - 

### Expected Behavior

Ensure that there is at least one step between the highest value and the diagram’s upper limit, improving readability and consistency.

DenisKoether:
### Related Issue

closes #310 
---

### DIFF
--- a/src/components/results/ChartComponent.wc.svelte
+++ b/src/components/results/ChartComponent.wc.svelte
@@ -44,6 +44,10 @@
         viewScales?: boolean;
         backgroundColor?: string[] | string;
         backgroundHoverColor?: string[];
+        xAxisStepSize?: number;
+        yAxisStepSize?: number;
+        xAxisPrecision?: number;
+        yAxisPrecision?: number;
     }
 
     let {
@@ -86,6 +90,10 @@
             "#80699b",
         ]),
         backgroundHoverColor = ["#aaaaaa"],
+        xAxisStepSize = 1,
+        yAxisStepSize = 1,
+        xAxisPrecision = 0,
+        yAxisPrecision = 0,
     }: Props = $props();
 
     // This is undefined if the lens options are not loaded yet
@@ -117,7 +125,7 @@
             labels: ["", "", "", ""],
             datasets: [
                 {
-                    data: [1, 1, 1, 1],
+                    data: [3, 1, 2, 5],
                     backgroundColor: ["#E6E6E6"],
                     backgroundHoverColor: ["#E6E6E6"],
                 },
@@ -157,6 +165,13 @@
                         display: true,
                         text: yAxisTitle,
                     },
+                    ticks: {
+                        stepSize: yAxisStepSize,
+                        callback: (value: number) => {
+                            return value.toFixed(yAxisPrecision);
+                        },
+                    },
+                    type: scaleType === "linear" ? "linear" : undefined,
                 },
                 x: {
                     display: viewScales,
@@ -167,6 +182,7 @@
                     ticks:
                         chartType === "bar"
                             ? {
+                                  stepSize: xAxisStepSize,
                                   callback: (val: string | number) => {
                                       if (indexAxis === "y")
                                           return val.toString();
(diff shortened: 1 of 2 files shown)


patrick-skowronek (changes requested):
There are two things wrong with this change. 

1. The default
Bildschirmfoto 2025-03-27 um 10 48 36
 values for flipped axis is not working 
2. Also on the flipped one, it does not render it anymore and I get these Error messages `scales.y.ticks.stepSize: 1 would result generating up to 440927 ticks. Limiting to 1000.`

patrick-skowronek:
Hey @timakro , I finished the PR could you please have a look at it

DenisKoether:
@patrickskowronekdkfz Patrick jetzt ist es doch wie vorher? stepsize und Nachkommastelle lässt sich nicht einstellen?

DenisKoether:
@patrickskowronekdkfz Was wäre wenn man den Chart nach eigenen Angaben anpassen möchte? z.B. Stepszie von 5 mit 2 Nachkommastellen. Dies ist so leider mit dem Fix nicht mehr möglich.

patrick-skowronek:
> @patrickskowronekdkfz Was wäre wenn man den Chart nach eigenen Angaben anpassen möchte? z.B. Stepszie von 5 mit 2 Nachkommastellen. Dies ist so leider mit dem Fix nicht mehr möglich.

I tested multiple options. Step size works with half values, like 0.5. However, when you have values that go over a specific threshold, chartjs will remove the precision. With the options you can set it more then the threshold, however it does not look good.

src/components/results/ChartComponent.wc.svelte
```diff
@@ -46,6 +44,9 @@
         viewScales?: boolean;
         backgroundColor?: string[] | string;
         backgroundHoverColor?: string[];
+        xAxisStepSize?: number;
+        yAxisStepSize?: number;
+        useIntegerPrecision?: boolean | undefined;
```
timakro: Remove this if it's not used

src/components/results/ChartComponent.wc.svelte
```diff
@@ -63,6 +64,8 @@
         groupRange = $bindable(0),
         groupingDivider = "",
         filterRegex = "",
+        xAxisStepSize = undefined,
+        yAxisStepSize = 1,
```
timakro: Maybe don't set a default and let Chart.js choose a sensible step size if it's not specified. Otherwise it may look wonky if the application doesn't explicitly set a step size:

image

timakro (changes requested):
There seems to be an issue with this one graph of the CCP explorer that needs to be fixed:

image

src/components/results/ChartComponent.wc.svelte
patrick-skowronek: I've thought about it, and you might be right. I don't think anyone will ever actually need to set a custom step size. That said, I do think it still makes sense for the pre-result — at least from my perspective.

src/components/results/ChartComponent.wc.svelte
patrick-skowronek: We currently let Chart.js determine the step size automatically — in most cases, it rounds to a reasonable next step. However, our users expect there to always be a visible step between the highest bar and the maximum value. I'll look into how we can ensure that.

patrick-skowronek:
> There seems to be an issue with this one graph of the CCP explorer that needs to be fixed:
> 
> image

I will look into it. Something about the Step size is not working correctly

timakro:
Also can you rebase please, this PR's diff includes a change  that has been merged 3 weeks ago (https://github.com/samply/lens/pull/279/commits/780ba7206f3f2a39b2ccddd26bbddeaa392f2ecb)

src/components/results/ChartComponent.wc.svelte
timakro: `useIntegerPrecision` is used nowhere in the code

src/components/results/ChartComponent.wc.svelte
timakro: Changing line 68 to `yAxisStepSize = undefined` should probably be enough

patrick-skowronek:
@timakro it should now be good, feel free to comment

timakro (approved):
image

Looking good!

patrick-skowronek (approved):

PR #302, opened 2025-03-31
chore: export missing types

### LINKED ISSUE #262: Export all return/parameter types of exported functions
E.g. `setCatalogue` takes a parameter of type `Catalogue` so that type should be exported. Typedoc shows warnings if a type export is missing, to see those run `npm run typedoc`.

timakro:
This PR exports types that are return types or argument types of exported functions

Closes #262


patrick-skowronek (changes requested):
Could you please also export the function `resolveAstSubCatagories` that we can use it in other projects

patrick-skowronek (approved):

Issue #306, opened 2025-04-01
[Feature]: Loading spinner


patrick-skowronek:
### General Summary

When a user triggers a query, the only visible change in the UI is in the result table. This can make it unclear whether the system is actively processing the request.

Solution: Add a loading spinner that remains visible until one of the following conditions is met:

All queried sites have either returned a result or reported a failure.

The query exceeds the designated time limit.

This will improve user feedback and make the query process more transparent.

### Is this Breaking?

No

PR #315, opened 2025-04-01
feat: added query loading spinner

patrick-skowronek:
### Description

This PR adds a loading spinner, which spins until the all sites have responded

### Related Issue

#306 

---

### DIFF
--- a/src/components/informational/QuerySpinner.wc.svelte
+++ b/src/components/informational/QuerySpinner.wc.svelte
@@ -8,9 +8,9 @@
     import { responseStore } from "../../stores/response";
     import type { ResponseStore } from "../../types/backend";
 
-    export let size = "20px";
+    let { size = "20px" } = $props();
 
-    let loading = false;
+    let loading = $state(false);
 
     window.addEventListener("emit-lens-query", function () {
         loading = true;
@@ -21,8 +21,8 @@
             if (value[1].status === "claimed") {
                 return;
             }
-            loading = false;
         }
+        loading = false;
     });
 </script>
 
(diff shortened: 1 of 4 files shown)


src/components/informational/QuerySpinner.wc.svelte
```diff
@@ -0,0 +1,51 @@
...
+    customElement={{
+        tag: "lens-query-spinner",
+    }}
+/>
+
+<script lang="ts">
+    import { responseStore } from "../../stores/response";
+    import type { ResponseStore } from "../../types/backend";
+
+    export let size = "20px";
```
timakro: https://svelte.dev/docs/svelte/v5-migration-guide#Reactivity-syntax-changes-export-let-$props

src/components/informational/QuerySpinner.wc.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+    let loading = false;
```
timakro: https://svelte.dev/docs/svelte/v5-migration-guide#Reactivity-syntax-changes-let-$state

timakro (changes requested):
It looks like Svelte 4 syntax still works in Svelte 5:

> You don’t have to migrate to the new syntax right away - Svelte 5 still supports the old Svelte 4 syntax, and you can mix and match components using the new syntax with components using the old and vice versa.

(from https://svelte.dev/docs/svelte/v5-migration-guide)

But we should still use the new syntax in new code.

src/components/informational/QuerySpinner.wc.svelte
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    responseStore.subscribe((s: ResponseStore) => {
+        for (const value of s) {
+            if (value[1].status === "claimed") {
+                return;
+            }
+            loading = false;
+        }
+    });
```
timakro: This logic means if any of the sites is not claimed, stop the spinner. Don't we want to wait for all of the sites to be not claimed until we stop the spinner?

timakro (approved):

PR #316, opened 2025-04-01
Upgraded Feat/catalog string type to v0.5.0-alpha

### LINKED ISSUE #223: [Feature]: String type for Lens catalog
### General Summary

Lens has a number of different allowed data types for the attributes in the search term catalog, e.g. "numeric", but so far, no "string" type. The GDI/Beacon and the ECDC/EHDS2 implementations of Lens requires the "string" type.

### Suggested Change

This new feature will allow users to enter arbitrary strings.

### Is this Breaking?

No

### Current Behavior

Currently, users cannot enter strings.

### New Behavior

The new functionality would allow users to enter strings.

DavidCroftDKFZ:
### Description
Upgraded Feat/catalog string type to v0.5.0-alpha

### How Has This Been Tested?

Locally using npm link to the lens2beacon repo


timakro:
@DavidCroftDKFZ I added a string field to the catalogue of the dev.ts page. You can test it by running `npm run dev`. It doesn't seem to work yet as I cannot add the string to the query:

image

DavidCroftDKFZ:
@timakro I tried your change to dev.ts, and it worked for me. What was the problem? Did the field not accept text? Or did the text not appear in the search bar? Note: you need to hit the enter/newline key to make it pass the string to the search bar.

timakro:
You are right. But could you also add an arrow button next to the input field? You can check how the other input components do it.

DavidCroftDKFZ:
OK, will do.

timakro:
I went ahead and did it in e53152cbe99f2bd5ef928e70a9f740a7bff89702 (didn't see your comment)

dev.ts
```diff
@@ -173,10 +173,18 @@ setCatalogue([
...
+        key: "sample-id",
+        name: "Sample ID",
+        type: "EQUALS",
+        system: "",
+    },
 ]);
 
 window.addEventListener("emit-lens-query", (event) => {
     const detail = (event as QueryEvent).detail;
+    console.log("AST:", JSON.stringify(detail.ast));
```
patrick-skowronek, line 187: Please don't do not commit consol.log

patrick-skowronek (changes requested):
Looks good to me, only one small thing.
Right now, it's set to 'equals,' but in some cases, 'contains' might make more sense. Something to consider for the future!

dev.ts
timakro: I added the `console.log` in `dev.ts` so we can see the AST that gets sent out. I think that's useful.

patrick-skowronek (approved):

Issue #265, opened 2025-03-21
[Feature]: Make Dateinput stacked instead of side by side


patrick-skowronek:
### General Summary

The Dateinput is wide, most of the time more wide then the catalogue. The design should be improved, with either stacked elements or a other solution.

<img width="399" alt="Image" src="https://github.com/user-attachments/assets/63daa49c-6523-4d2e-8f66-c95130f3077d" />

### Is this Breaking?

No

timakro:
We decided to keep them side by side https://github.com/samply/lens/pull/345#issuecomment-2796033597

PR #345, opened 2025-04-09
chore: rework number and date inputs

### LINKED ISSUE #159: [Feature]: set min/max date default in catalog
### General Summary

Currently, there's no easy way to set a min/max date in the catalog. Adding this feature would allow users to filter catalog entries by date range, improving usability and navigation.

### Suggested Change

- Implement date pickers for setting min/max dates.
- Update catalog filtering to respect the date range.

image

### Is this Breaking?

No

### Current Behavior

none

### New Behavior

none

### Detailed Description

_No response_

### LINKED ISSUE #264: [Feature]: Disable Number and Date add Button if from and to field are empty
### General Summary

Make the input more strict when both are empty disable the button

### Is this Breaking?

No

### LINKED ISSUE #307: [Bug]: Adding invalid options from catalogue
### General Summary

Users can currently enter invalid values in the search bar when using date or number inputs from the catalog. This should not be possible.

### Current Behavior

A number input with 10 on the low limit and -1 on the high limit would enter a invalid chip to the searchbar

### Expected Behavior

The input button should be disabled

timakro:
This PR restyles the number and date inputs and adds error handling:

image
---
image

Due to the error handling this PR closes #264 and closes #159 and closes #307.

I don't think stacking the date inputs as suggested in #265 is a good idea. This PR removes the "from" and "to" texts so it might be a little bit less wide than before.

I also added optional units to the numer inputs as I figured we might need them at some point.

Things that are still to do:

- [x] Translate custom error messages
- [x] Decide if we want to put styles inside the components `.svelte` files
- [x] Decide if we want to use the circular buttons for all catalogue items


patrick-skowronek:
Here are my thoughts on the changes:

1. I like the addition of the number input type—it makes it clearer to the user what should be entered. However, I’d prefer placing it only at the end, maybe something like . 
Bildschirmfoto 2025-04-11 um 08 38 16
Since we are size-constrained, I’d always aim to keep it as compact as possible.

2. I prefer the plus button over the arrow. I’d suggest using it for all inputs.

3. Removing the 'from' and 'to' text is a good improvement. Similar to point 1, I’d group all min/max inputs into one bracket, as it saves a lot of space. I tried stacking them but couldn’t make it work. We can probably close #265—I measured the sizes, and we reduced the width by about 100px, which looks much better.

4. I think providing feedback when a user enters an invalid value is a great idea. The error messages should be translated and improved for better clarity.

5. Regarding styling, I prefer keeping it within the component, as discussed. However, does it make sense to place it at the bottom part? Is there a standard from svelte where to put the html, script and style part. I think we should maintain a consistent order across all files.

timakro:
PR is ready for review. I've opted for the separated input design because it clearly shows which input has focus with the outline.

I added these changes:
* Add docs about the new component-scoped styling convention
* Move some styles from separate CSS files into components
* Add warning in v0.5.1 migration guide that style overwrites for catalogue components will likely break
* Add translations for number/date input error messages
* Use the circular plus button everywhere

timakro:
Removed AST guide from book TOC because I don't want an empty page linked there - I first tried to comment it out but Prettier absolutely did not like it...

patrick-skowronek (approved):

PR #348, opened 2025-04-11
chore: added some styling to the demo

patrick-skowronek:
### Description

Some styling in the for the demo

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,27 +20,23 @@
             }
 
             .container {
-                text-align: left;
+                display: flex;
+            }
+
+            .searchbar {
+                display: grid;
+                grid-template-columns: 25fr 1fr 1fr;
+                align-items: center;
             }
 
             .box {
-                display: inline-block;
                 padding: 10px;
-                margin-right: 20px;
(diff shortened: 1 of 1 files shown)


timakro:
I think we shouldn't try to make the dev page pretty. If we start going down that route we will have to add a lot more styles to make it palatable to an end user and that shouldn't be the goal of the dev page.

As an exception I think layout changes for usability are a good idea. If you want to place the catalogue next to the graphs to use the screen space better and prevent it form shifting the graphs I think that is a good idea. Your layout has some issues though: the "expand tree" button shifts (1) when you open/close the catalogue and (2) when the search spinner becomes visible. Then again the catalogue should be open by default anyways like in ccp explorer so just doing that would be sufficient right now to make it usable/testable imo.

Regarding the implementation I would suggest you use flexbox to place elements side by side. I haven't seen clearfix in years :D

patrick-skowronek:
I think my changes are not pretty 😅 and it should be as simple as possible. But I would add those and not more

timakro (approved):

Issue #308, opened 2025-04-01
[Feature]: Reset diagrams


patrick-skowronek:
### General Summary

A user requested the ability to reset all diagrams. Currently, there is no way to do this efficiently. Solution: Add a Reset button component that:

1. Clears the result store.

2. Automatically resets all diagrams to their default skeleton.

This will allow users to quickly return to a clean state without manually resetting each diagram. The reset function should be exported as well.

### Is this Breaking?

No

paolaklein:
https://github.com/samply/lens/tree/reset-diagrams

https://github.com/user-attachments/assets/b3e1675f-c829-4055-b0e9-1db054e4bacf

timakro:
Closed by #355

PR #355, opened 2025-04-16
Reset diagrams

paolaklein:
### Description

Guide on how a button could look like.
service for the reset function + export
(and the functions to clear: query, responseStore and Charts)

### Related Issue

https://github.com/samply/lens/issues/308

---

### DIFF
--- a/book/src/guide/reset-button.md
+++ b/book/src/guide/reset-button.md
@@ -1,24 +1,22 @@
-
 This is how you would implement a reset button using the resetDiagrams functionn (service: reset.ts)
 
 ```svelte
-<!-- ToDo Change "Reset" to be dynamic (translated)-->
-    <button class="reset-button" onclick={resetDiagrams}>Reset</button>
+<!-- example for a simple button-->
+<button class="reset-button" onclick={resetDiagrams}>Reset</button>
 
-<!-- ToDo Move Style ? -->
+<!-- styling to fit with the rest/other buttons -->
 <style>
     button.reset-button {
-    background-color: var(--button-background-color);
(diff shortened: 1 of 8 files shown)


book/src/guide/reset-button.md
```diff
@@ -0,0 +1,24 @@
+
+This is how you would implement a reset button using the resetDiagrams functionn (service: reset.ts)
+
+```svelte
+<!-- ToDo Change "Reset" to be dynamic (translated)-->
```
patrick-skowronek: Please remove the todo

src/components/results/ChartComponent.wc.svelte
```diff
@@ -474,13 +474,32 @@
...
+     */
+    export function resetChart() {
+        if (!chart) return;
+
+        chart.data.labels = ["", "", "", ""];
+        chart.data.datasets = [
+            {
+                data: [1, 1, 1, 1],
+                backgroundColor: ["#E6E6E6"],
+                hoverBackgroundColor: ["#E6E6E6"], // or remove if Chart.js doesn't recognize it
```
patrick-skowronek: Why would Chart.js not recognize the Color?

patrick-skowronek (changes requested):
Please fix your linting.

timakro:
`resetDiagrams()` is supposed to be called by the application? Then it needs to be exported so the application can see it.

Another suggestion: Why not create a reset button component that the application can place if wanted that calls `resetDiagrams()` internally?

patrick-skowronek (approved):

PR #392, opened 2025-05-15
Feat/paginator size

patrick-skowronek:
### Description

Adds an option to dynamically change the result page size.

### DIFF
--- a/src/components/results/ResultTableComponent.wc.svelte
+++ b/src/components/results/ResultTableComponent.wc.svelte
@@ -122,9 +122,11 @@
     interface Props {
         title?: string;
         pageSize?: number;
+        showPageSize?: boolean;
     }
 
-    let { title = "", pageSize = 10 }: Props = $props();
+    let { title = "", pageSize = 10, showPageSize = false }: Props = $props();
+
     let activePage = $state(1);
     let sortColumnIndex = $state(0);
     let sortAscending = $state(true);
@@ -169,6 +171,20 @@
             sortAscending = !sortAscending;
         }
     }
+
+    let pageSizeOptions: { size: number; text: string }[] = [
+        { size: 10, text: `10` },
+        { size: 25, text: `25` },
+        { size: 50, text: `50` },
+    ];
+
+    if (pageSize != 10 && pageSize != 25 && pageSize != 50) {
+        pageSizeOptions.push({ size: pageSize, text: String(pageSize) });
+    }
+
+    let selected = $state();
+
+    console.log(showPageSize);
 </script>
 
 <h4 part="result-table-title">{title}</h4>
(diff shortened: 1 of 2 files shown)


src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -122,9 +122,18 @@
...
         pageSize?: number;
+        showPageSize?: boolean;
     }
 
-    let { title = "", pageSize = 10 }: Props = $props();
+    /**
+     *  Configuration options for the result table.
+     * title:  The title to be displayed in the table header.
+     * pagesize: The number of rows to display per page. Defaults to 10.
+     * showPageSize: Whether to display a select box for changing the pagesize. If you set the pagesize otherthen 10,25,50 it will added as an option.
```
timakro: Better move this to the individual prop definitions in the interface. That way you will get IntelliSense for the props when using the component.

src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -122,9 +122,18 @@
     interface Props {
         title?: string;
         pageSize?: number;
+        showPageSize?: boolean;
```
timakro: Maybe call it `showPageSizeSwitcher`

timakro (commented):
I think allowing apps to change the page size is a good idea but I'm not sure about the drop down that allows the user to change the page size. Usually the number of sites is static per app so apps should be able to choose a page size that works well for their layout and number of sites. Do you have some app/use case where you need this?

patrick-skowronek:
> I think allowing apps to change the page size is a good idea but I'm not sure about the drop down that allows the user to change the page size. Usually the number of sites is static per app so apps should be able to choose a page size that works well for their layout and number of sites. Do you have some app/use case where you need this?

I agree with you. Changing the page size is already a feature. However, this feature is requested by the BBMRI-ERIC Sample Locator, where users want to view more sites simultaneously. It’s true that this change will significantly alter the layout. While it’s understandable that users might prefer the default of 10 pages, I don’t believe adding a button to display all results is a necessary feature.

src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -119,12 +119,24 @@
...
 
+    /**
+     *  Configuration options for the result table.
+     */
     interface Props {
+        /** title:  The title to be displayed in the table header. */
         title?: string;
+        /** pagesize: The number of rows to display per page. Defaults to 10. */
         pageSize?: number;
+        /** pageSizeSwitcher: Whether to display a select box for changing the pagesize. If you set the pagesize otherthen 10,25,50 it will added as an option. */
```
timakro: Repeating the name of the variable in the doc string is weird

src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -169,6 +181,19 @@
...
+        { size: 25, text: `25` },
+        { size: 50, text: `50` },
+    ];
+
+    if (pageSize != 10 && pageSize != 25 && pageSize != 50) {
+        pageSizeOptions.push({ size: pageSize, text: String(pageSize) });
+        pageSizeOptions.sort((a, b) => a.size - b.size);
+    }
+
+    let selected = $state(pageSizeOptions[0]);
```
timakro: could use a better name than `selected`

timakro (approved):

PR #437, opened 2025-06-17
fix: solve some layout issues

timakro:
I think we broke the chart layout in 0.5.1, from that version onwards charts only grow but never shrink.

There is this comment in a Chart.JS issue https://github.com/chartjs/Chart.js/issues/4156#issuecomment-295180128 that says you have to set `min-width: 0` on the Chart container so it can shrink. This is also what I am doing in the organoid dashboard. I'm a bit confused because I don't think we did this in 0.5.0 and graphs used to shrink anyways.

This PR sets `min-width: 0` on a dedicated chart container `div` so charts can shrink again. I've also fixed catalogue items so they wrap when they are too long. I also had to set `height: 100%` on the chart wrapper so it takes the entire vertical space that is reserved by the application for the chart. Some deal here, I am not sure why this wasn't needed before and is needed now.

### DIFF
--- a/src/components/results/ChartComponent.wc.svelte
+++ b/src/components/results/ChartComponent.wc.svelte
@@ -666,19 +666,23 @@
         </div>
     {/if}
 
-    <canvas
-        part="chart-canvas"
-        bind:this={canvas}
-        id="chart"
-        onclick={handleClickOnStratifier}
-    ></canvas>
+    <!-- For responsive charts, Charts.js requires a dedicated container for each canvas: https://www.chartjs.org/docs/latest/configuration/responsive.html#important-note -->
+    <!-- The container requires `min-width: 0` or it won't shrink: https://github.com/chartjs/Chart.js/issues/4156#issuecomment-295180128 -->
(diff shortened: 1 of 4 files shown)


src/components/results/ChartComponent.wc.svelte
```diff
@@ -666,19 +666,23 @@
...
 
-    <canvas
-        part="chart-canvas"
-        bind:this={canvas}
-        id="chart"
-        onclick={handleClickOnStratifier}
-    ></canvas>
+    <!-- For responsive charts, Charts.js requires a dedicated container for each canvas: https://www.chartjs.org/docs/latest/configuration/responsive.html#important-note -->
+    <!-- The container requires `min-width: 0` or it won't shrink: https://github.com/chartjs/Chart.js/issues/4156#issuecomment-295180128 -->
+    <div style="min-width: 0;">
```
patrick-skowronek: Move to class

patrick-skowronek (changes requested):

patrick-skowronek (approved):

PR #460, opened 2025-06-30
feat: rename classes to backends

patrick-skowronek:
### Description

Better Naming for Backends


timakro (approved):

timakro:
Approved but seems like CI fails

PR #466, opened 2025-07-01
feat: improve explain query button for all cases

patrick-skowronek:
### Description

This PR changes the output of the query explain button, especially the between part

### DIFF
--- a/book/src/releases/v0.6.0.md
+++ b/book/src/releases/v0.6.0.md
@@ -1,0 +1,21 @@
+# Version 0.6.0
+
+## Migration guide
+
+### QueryExplainButton
+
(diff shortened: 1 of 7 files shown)


timakro:
This is wrong:

image

I guess we are still using 0 to represent no minimum/no maximum. Would really like to see this changed.

patrick-skowronek:
> This is wrong:
> 
> image
> 
> I guess we are still using 0 to represent no minimum/no maximum. Would really like to see this changed.

You are right, this should be different. I think it should be undefined in the lens and not be there in the AST. I'm a bit worried that this will break something.

timakro (approved):

timakro:
I guess the one-sided interval thing is still broken though

patrick-skowronek:
> I guess the one-sided interval thing is still broken though

Yes, I wouldn't mind to merge it and make a new PR for the restructuring of Query/AST

PR #471, opened 2025-07-03
feat: add hint to singleselect when description is provided

### LINKED ISSUE #465: [Feature]: Display description of each item in singleselect with hover
### General Summary

We got the feedback to display the description not only in the autocomplete of a catalogue item, but instead with a hover effect in the catalogue component.

### Is this Breaking?

No

patrick-skowronek:
### Description

This PR adds a hint to a singleselect item if a description is provided 

closes #465

### DIFF
--- a/dev.svelte
+++ b/dev.svelte
@@ -256,7 +256,8 @@
                                 {
                                     key: "urn:dktk:code:3:2",
                                     name: "Gliom - Grad I",
-                                    description: "",
+                                    description:
+                                        "Diagnosis D43.% and Morph 9383/1,9384/1",
                                     aggregatedValue: [
                                         [
                                             {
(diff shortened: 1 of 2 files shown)


timakro:
image

Does this look too noisy?

Also we should probably remove these useless descriptions in CCP explorer.

timakro (approved):

patrick-skowronek:
> image
> 
> Does this look too noisy?
> 
> Also we should probably remove these useless descriptions in CCP explorer.

In general I would say no, in the case of the CCP-Explorer yes. I would remove any description that does not improve the understanding of elements. We should mention that in the docs

timakro:
Fine to merge from my side

PR #477, opened 2025-07-04
One sided range query representation

timakro:
In BBMRI sample locator one-sided ranges will cause the query to fail because focus does not support them. I guess this is better behavior than silently putting 0 in place of the missing value.


patrick-skowronek (approved):
We should document the query somewhere

PR #481, opened 2025-07-04
Query in url

### LINKED ISSUE #467: Have query in the URL
The URL should always reflect the current search query. When the user opens a page with a query in the URL it should be loaded into the search bar.

We can have this feature enabled by default and allow to turn it off via lens options.

timakro:
Some text encoding stuff is required to handle non-ASCII chars correctly.

Closes #467

### DIFF
--- a/src/stores/query.ts
+++ b/src/stores/query.ts
@@ -7,17 +7,23 @@ import { writable, get } from "svelte/store";
 import { v4 as uuidv4 } from "uuid";
 import type { Category, Criteria } from "../types/catalogue";
 import { lensOptions } from "../stores/options";
+import { showErrorToast } from "./toasts";
 
 export const queryStore = writable<QueryItem[][]>([[]]);
 
 const encodedQuery = new URLSearchParams(window.location.search).get("query");
 if (encodedQuery !== null) {
-    const query = JSON.parse(
-        new TextDecoder().decode(
-            Uint8Array.from(atob(encodedQuery), (c) => c.charCodeAt(0)),
-        ),
-    );
-    queryStore.set(query);
+    try {
+        const query = JSON.parse(
+            new TextDecoder().decode(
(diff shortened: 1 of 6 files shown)


src/stores/query.ts
```diff
@@ -6,20 +6,18 @@ import type { QueryItem, QueryValue } from "../types/queryData";
...
-const urlParams: URLSearchParams = new URLSearchParams(window.location.search);
-const queryParam: string | null = urlParams.get("query");
-
-if (queryParam !== null) {
-    const queryParamDecoded: QueryItem[][] = JSON.parse(atob(queryParam));
-    queryStore.set(queryParamDecoded);
+const encodedQuery = new URLSearchParams(window.location.search).get("query");
+if (encodedQuery !== null) {
+    const query = JSON.parse(
+        new TextDecoder().decode(
```
patrick-skowronek: If the String is not a Base64 String this will fail without any error message to the user. I would suggest a try and catch, if the string is not parseable, the query store should be set empty and a error should be shown.

patrick-skowronek (changes requested):

timakro:
I had to move the code to an `onMount` so the lens options are loaded (for the translation language) when it runs. I chose `SearchBarComponent` because it makes the most sense I think.

We could also try `setTimeout(() => {/* code here */}, 0);` and keep the code in `query.ts` if you prefer that. Not sure what is less ugly ^^

patrick-skowronek (approved):

patrick-skowronek:
> I had to move the code to an `onMount` so the lens options are loaded (for the translation language) when it runs. I chose `SearchBarComponent` because it makes the most sense I think.
> 
> We could also try `setTimeout(() => {/* code here */}, 0);` and keep the code in `query.ts` if you prefer that. Not sure what is less ugly ^^

I think `onMount`  works for now, I would start with setTimeout. We would need something like a lifecycle, but that is very overkill 😅

PR #484, opened 2025-07-07
chore: upgraded vite to version 7

patrick-skowronek:
### Description

Upgrade to vite 7, should be faster

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
                 "typedoc": "^0.28.1",
                 "typescript": "^5.8.2",
                 "typescript-eslint": "^8.26.1",
-                "vite": "^6.2.1",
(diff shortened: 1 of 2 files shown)


timakro:
CI fails?

patrick-skowronek:
> CI fails?

Yes, the dependencies are not updated yet

patrick-skowronek:
https://github.com/sveltejs/vite-plugin-svelte/issues/1150

timakro:
I think we should wait for a full release of vite-plugin-svelte before we upgrade. What do you think?

patrick-skowronek:
> I think we should wait for a full release of vite-plugin-svelte before we upgrade. What do you think?

I agree, the difference it not that big. I would use this PR for release tracking

PR #485, opened 2025-07-07
Use hardcoded icons

### LINKED ISSUE #482: Remove icon URLs from lens options
Remove these from the lens options:

```
export type IconUrls = {
    /** Icon for an info button used in various places */
    infoUrl?: string;
    /** Icon for a delete button used in various places */
    deleteUrl?: string;
    /** Icon for the button that adds items from the catalogue to the search bar */
    addIconUrl?: string;
    /** Icon for the button that expands and collapses items in the catalogue */
    toggleIconUrl?: string;
};
```

Motivation: You should get something that looks nice by default without having to copy the image files into your app and putting the URLs into the options.

Two options
1. Remove the possibility to customize icons and just paste SVG into the HTML
2. Find some way to specify image via CSS so apps can overwrite the images just like they would overwrite any styles. For example:
    ```
    element {
      background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='100' height='100'><circle cx='50' cy='50' r='40' stroke='green' fill='yellow' stroke-width='4'/></svg>");
    }
    ```

timakro:
image
image

I also took the opportunity to remove the global styles for the various info buttons and instead make the info button component tweakable using `alignDialogue?: "left" | "right" | "center"` and `dialogueMaxWidth?: string` props. This should hopefully allow to configure the buttons such that dialogue texts are always in the viewport.

Closes #482


book/src/releases/v0.6.0.md
```diff
@@ -78,3 +78,18 @@ The URL now always updates to stay in sync with the current query so can share a
 ```json
 "autoUpdateQueryInUrl": false
 ```
+
+### Removed iconOptions
```
patrick-skowronek, line 82: Please also add the changes to the lens-explain-query-button

src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -5,13 +5,22 @@
...
+        buttonSize?: string;
+        alignDialogue?: "left" | "right" | "center";
+        dialogueMaxWidth?: string;
+        /** Info button in search bar is white and orange on hover */
+        inSearchBar?: boolean;
     }
 
-    let { message = $bindable([]) }: Props = $props();
+    let {
+        message = "",
```
patrick-skowronek, line 18: I don't mind the options. I was pretty optimistic to override it with the part mechanism. But in the end the user would properly use this instead.

patrick-skowronek (changes requested):
Just add the documentation.

book/src/releases/v0.6.0.md
timakro: The `inSearchBar` prop isn't really intended to be used by applications. So I don't think we should even document it.

src/components/buttons/InfoButtonComponent.wc.svelte
timakro: At some point we should change the info button to use the [popover API](https://developer.chrome.com/blog/introducing-popover-api/#the-difference-between-a-popover-and-a-dialog). Browsers can do popovers natively without need for javascript but it only really makes sense in combination with CSS anchor positioning and [Firefox is still lacking support](https://caniuse.com/css-anchor-positioning). It's really cool it allows you to write easily in CSS: "place this dialogue below the button but if it would escape the viewport move it over so it's visible"

patrick-skowronek: That would be really nice

patrick-skowronek (approved):

Issue #491, opened 2025-07-08
Remove AST2CQL generation in Lib


patrick-skowronek:
The Lib should not contain the dktk ast2cql translator. I would suggest to remove it and add documentation

timakro:
Moving the actual CQL generation is easy. But the project manager needs the CQL generation so we should probably move that too and that is not so easy. Will have to look more into this

patrick-skowronek:
Yes, you're right. I looking also into it.

PR #492, opened 2025-07-08
Remove lens options component

### LINKED ISSUE #428: Remove <lens-options> component
The <lens-options> component is already completely optional to use and everything it does can be done with `setOptions`, `setCatalogue` and `setMeasures` instead. I also changed BBMRI to use the new functions which means we no longer need the async loading logic [see the diff](https://github.com/samply/bbmri-sample-locator/commit/658b61ad11004cc28118aea8c950df6c45f3dc0f?w=1). We should encourage everyone to use the new functions so the async loading `{#await}` blocks are not needed.

timakro:
Closes #428


src/components/Options.wc.svelte
```diff
@@ -1,40 +0,0 @@
...
-    }: Props = $props();
-
-    $effect(() => {
-        setOptions(JSON.parse(optionsJSON));
-    });
-
-    $effect(() => {
-        setCatalogue(JSON.parse(catalogueJSON));
-    });
-    $effect(() => {
```
patrick-skowronek, line 37: This will break DKTK cql generation, therefore we need to do #491

patrick-skowronek (approved):

src/components/Options.wc.svelte
timakro: Lens exports `setMeasures` right now. Maybe we should remove the measure store in lens completely and have the app deal with that I am not sure.

PR #507, opened 2025-07-11
fix: close event source when server closes connection

timakro:
Removed this when rewriting the spot class to a function. Turns out this is needed or the same results will keep coming in and refreshing the graphs again and agin.


patrick-skowronek (approved):
Nice

PR #508, opened 2025-07-14
Create ovis.ts, add tests, remove addStratifierToQueryAPI

### LINKED ISSUE #219: [Bug]: Empty query bug (Ovis)
### General Summary

As soon as you empty an already filled query again (i.e. remove all conditions) OVis can no longer search properly. I suspect this is currently related to the unintentional AST change (discussed last week in the Lens chat).

### Expected Behavior

-

### Current Behavior

Ovis Query auswählen. Beim deselect kommt folgender Fehler:

lens.js?v=512874ff:1788  Uncaught (in promise) TypeError: Cannot read properties of null (reading 'children')
    at Hn (lens.js?v=512874ff:1788:12)
    at r.<anonymous> (lens.js?v=512874ff:18104:11)
    at QuicktoolsActiveFilters.svelte:26:33

### Possible Solution

-

### Steps to Reproduce

?

### Context (Environment)

-

### Detailed Description

-

### LINKED ISSUE #493: [Bug]: addStratifierToQuery doesn't handle numeric range correctly (OVIS)
### General Summary

The maximum value isn't parsed correctly and is always set to 1.

### Current Behavior

-

### Expected Behavior

-

timakro:
Closes #219 and closes #493


book/src/releases/v0.6.0.md
```diff
@@ -98,17 +98,18 @@ You can also remove the icon image files if your app does not use them anywhere
...
+| `getCatalogueAPI`                                   | Apps control the catalogue so there should be no need to read it from Lens.                                                                                                                                             |
+| `setQueryStoreAPI`                                  | [`setQueryStore`](https://samply.github.io/lens/docs/functions/setQueryStore.html)                                                                                                                                      |
+| `setQueryStoreFromAstAPI`                           | Only OVIS uses this API and we have provided a replacement for them but use of this function in new code is discouraged.                                                                                                |
+| `addStratifierToQueryAPI`                           | [`addItemToActiveQueryGroup`](https://samply.github.io/lens/docs/functions/addItemToActiveQueryGroup.html)                                                                                                              |
```
timakro, line 111: Diff a bit hard to read, only touched these two lines.

src/backends/spot.ts
```diff
@@ -23,7 +25,7 @@ export async function createBeamTask(
     resultCallback: (result: BeamResult) => void,
 ): Promise<void> {
     url = url.endsWith("/") ? url : url + "/";
-    const id = crypto.randomUUID();
+    const id = uuidv4();
```
patrick-skowronek, line 28: Did we change this like a long time ago?

patrick-skowronek (changes requested):
Can we call the ovis file differently? I know we don't want to keep it for long, but I would argue for a better name.

src/backends/spot.ts
timakro: Apart from `crypto.randomUUID()` only working in secure contexts which can be a bit annoying they are identical. Just changed it for consistency because most places where we need a UUID already use the `uuid` npm package.

timakro:
> Can we call the ovis file differently? I know we don't want to keep it for long, but I would argue for a better name.

It only contains AST to query translation right now, so what about calling it `ast-to-query.ts`?

src/backends/spot.ts
patrick-skowronek: I know we had this before. Didn't we changed it everywhere?

patrick-skowronek:
> > Can we call the ovis file differently? I know we don't want to keep it for long, but I would argue for a better name.
> 
> It only contains AST to query translation right now, so what about calling it `ast-to-query.ts`?

Sounds good, I prefer this over ovis

patrick-skowronek (approved):

PR #519, opened 2025-07-15
Clean up types

timakro:
Remove `legacyUpdateResponseStore` and plenty of unused types.

Also remove blaze class.


patrick-skowronek (approved):
Looks good for now. I would strongly argue to add to the other things the Fhir prefix. We will do that in another PR

PR #525, opened 2025-07-17
chore: remove catalogueKeyToResponseKeyMap option and unify the interface

### LINKED ISSUE #521: [Feature]: Remove catalogueKeyToResponseKeyMap from options
### General Summary

Currently _catalogueKeyToResponseKeyMap_ is only used by the ChartsComponent and should be replaced by a simple Prop.

### Is this Breaking?

Yes

patrick-skowronek:
### Description

This PR does remove the catalogueKeyToResponseMap from the options. Lens Chart was the only component relying on it. 
BREAKING CHANGE: a lens chart now needs a dataKey

Closes #521

### DIFF
--- a/src/components/results/ChartComponent.wc.svelte
+++ b/src/components/results/ChartComponent.wc.svelte
@@ -27,7 +27,6 @@
 
     interface Props {
         title?: string; // e.g. 'Gender Distribution'
-        catalogueGroupCode?: string; // e.g. "gender"
         indexAxis?: string;
         xAxisTitle?: string;
         yAxisTitle?: string;
@@ -36,6 +35,7 @@
         displayLegends?: boolean;
         chartType?: keyof ChartTypeRegistry;
         scaleType?: string;
+        dataKey: string;
         perSite?: boolean;
         groupRange?: number;
         groupingDivider?: string;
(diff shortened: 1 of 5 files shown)


timakro:
Question remains how we make CCP explorer and other locators work without catalogueKeyToResponseKeyMap. Maybe the most sensible option is to just match everything to the FHIR store? So change the keys in the catalogue to match the FHIR store e.g. change `gender` to `Gender` in the catalogue. Let's have a chat about this.

timakro:
Okay let's go the nuclear route and remove catalogueKeyToResponseKeyMap and use the same keys everywhere (catalogue, AST, CQL generation, Lens result format, the `dataKey` prop of the `<lens-chart>` component). I think we need to be more clear in the migration guide what apps have to change because right now it sounds like everything will just work and that is not true?

src/components/results/ChartComponent.wc.svelte
```diff
@@ -196,12 +189,12 @@
 
     const accumulateValues = (
         valuesToAccumulate: string[],
-        catalogueGroupCode: string,
+        datakey: string,
```
timakro: Can you remove this parameter and use `dataKey` directly?

timakro:
What needs to go into migration guide:
* Remove `catalogueKeyToResponseKeyMap` from Lens options
* When using `<lens-chart>` change  `catalogueGroupCode` prop to `dataKey`
* Adjust CQL measures

PR #527, opened 2025-07-17
chore: reload page if SSO session expired

timakro:
Couldn't test this because my test environment does not include SSO. I would suggest we add this to the first release candidate of 0.6.0 and test if it works in the test CCP explorer.

It's kind of weird to add this to Lens since we don't even know if an application will be behind SSO but I think it's still better than having to add it to every application. What do you think?

### DIFF
--- a/src/backends/spot.ts
+++ b/src/backends/spot.ts
@@ -41,6 +41,12 @@ export async function querySpot(
     });
 
     if (!response.ok) {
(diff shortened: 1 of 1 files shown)


patrick-skowronek:
I'm not sure if this is the best UX. I agree we should build the test version and see how it feels.

patrick-skowronek (approved):

PR #554, opened 2025-07-28
chore: reload page when SSO session expired (second attempt)

timakro:


patrick-skowronek (approved):
Could work. I'm still somewhat cautious that the library reloads the page.

timakro:
Especially now that the code intercepts any redirect I get your concern. But if you don't do it in the library I feel like no app will care to handle this case because it rarely happens but is still annoying.

PR #566, opened 2025-07-31
docs: remove suggestion to fetch catalogue via REST call

timakro:
We should stick to the recommendation of importing the options/catalogue JSON directly instead of fetching it via REST call.


patrick-skowronek (approved):

patrick-skowronek:
I agree it is somewhat off in the docs. I should be clear with setCatalogue

PR #587, opened 2025-08-12
Fixes to the documentation in the mdbook.

mmwaikar:
### Description

The following documentation fixes are dine -

- The code snippet for the file `.vscode/settings.json`, should have the enclosing `{}` braces
- The `Test environment` should mention creating `./config/options-test.json` file
- Remove the word `the` from the text - `you could handle the these variable`
- *Query Store* - Figure shows blood group `A+` whereas the text mentions `A-`
- *Showing results* - typo `stratier (e.g. gender)`?
- *Chart* - typo `setable` instead of `are set`, `Non` instead of `None`
- *Result Summary* - where to set `resultSummaryOptions`?
- *Result Table* - where to set `tableOptions`?

### DIFF
--- a/book/src/components/resultsummary.md
+++ b/book/src/components/resultsummary.md
@@ -6,7 +6,7 @@ The `lens-result-summary` component displays a compact summary of result metrics
 
 ## Example
 
-To use it, define the configuration in the Lens options and include the component in your HTML:
+To use the component, define the configuration in the Lens [resultSummaryOptions](https://samply.github.io/lens/docs/types/ResultSummaryOptions.html) (in the `src/config/options.json` file as described in the [Options and catalogue](../guide/new-app.md#options-and-catalogue) section):
 
 ```json
 "resultSummaryOptions": {
@@ -19,14 +19,30 @@ To use it, define the configuration in the Lens options and include the componen
         }
     ]
 }
+```
 
+The `resultSummaryOptions` could also be set in JavaScript like this:
+
+```JavaScript
+let resultSummaryOptions = {
+  title: "Results",
+  infoButtonText: "This is a tooltip",
+  dataTypes: [
+    {
+      title: "Patients",
+      dataKey: "patients",
+    },
+    ],
+};
+
+options.resultSummaryOptions = resultSummaryOptions;
 ```
 
 ---
 
 ## Usage
 
-The component doesn't use any props and is only set via the [options](https://samply.github.io/lens/docs/types/ResultSummaryOptions.html)
+The component (which doesn't use any props) can then be included in your HTML:
 
 ```svelte
 <lens-result-summary></lens-result-summary>
--- a/book/src/components/resultsummary.md
+++ b/book/src/components/resultsummary.md
@@ -6,7 +6,7 @@ The `lens-result-summary` component displays a compact summary of result metrics
 
 ## Example
 
-To use the component, define the configuration in the Lens [resultSummaryOptions](https://samply.github.io/lens/docs/types/ResultSummaryOptions.html) (in the `src/config/options.json` file as described in the [Options and catalogue](../guide/new-app.md#options-and-catalogue) section):
+To use the component, define the configuration in the Lens [resultSummaryOptions](https://samply.github.io/lens/docs/types/ResultSummaryOptions.html) (as described in the [Options and catalogue](../guide/new-app.md#options-and-catalogue) section):
 
 ```json
 "resultSummaryOptions": {
@@ -19,23 +19,6 @@ To use the component, define the configuration in the Lens [resultSummaryOptions
         }
     ]
 }
-```
-
-The `resultSummaryOptions` could also be set in JavaScript like this:
-
-```JavaScript
-let resultSummaryOptions = {
-  title: "Results",
-  infoButtonText: "This is a tooltip",
-  dataTypes: [
-    {
-      title: "Patients",
-      dataKey: "patients",
-    },
-    ],
-};
-
-options.resultSummaryOptions = resultSummaryOptions;
 ```
 
 ---
--- a/book/src/components/resulttable.md
+++ b/book/src/components/resulttable.md
@@ -33,7 +33,7 @@ Each row includes a checkbox to select that data source for a data request. The 
 
 ## Example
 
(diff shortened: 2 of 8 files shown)


book/src/components/resultsummary.md
```diff
@@ -6,7 +6,7 @@ The `lens-result-summary` component displays a compact summary of result metrics
 
 ## Example
 
-To use it, define the configuration in the Lens options and include the component in your HTML:
+To use the component, define the configuration in the Lens [resultSummaryOptions](https://samply.github.io/lens/docs/types/ResultSummaryOptions.html) (in the `src/config/options.json` file as described in the [Options and catalogue](../guide/new-app.md#options-and-catalogue) section):
```
timakro: Maybe we don't want to be that concrete. `src/config/options.json` is the conventional name for the file but it's up to the application to choose. Or an app could decide to define their options in TypeScript instead of JSON.

book/src/components/resultsummary.md
```diff
@@ -19,14 +19,30 @@ To use it, define the configuration in the Lens options and include the componen
...
+  infoButtonText: "This is a tooltip",
+  dataTypes: [
+    {
+      title: "Patients",
+      dataKey: "patients",
+    },
+    ],
+};
+
+options.resultSummaryOptions = resultSummaryOptions;
```
timakro: I think we shouldn't start duplicating option examples in TypeScript. JSON files are the convention and if someone wants to do something special in TypeScript they can easily figure that out themselves.

book/src/guide/new-app.md
```diff
@@ -175,7 +173,9 @@ It is a common requirement to load different options in test and production. You
 - `PUBLIC_ENVIRONMENT`: Accepts the name of the environment, e.g. `production` or `test`
 - `PUBLIC_SPOT_URL`: Overwrites the URL of the [Spot](https://github.com/samply/spot) backend that your application queries
 
-For example you could handle the these variable as follows:
+Just like you created a JSON file `src/config/options.json` containing the empty object `{}`, create a `src/config/options-test.json` containing the empty object `{}`.
+
+Now you can handle these variables as follows:
```
timakro: The phrasing "create file containing empty object" seemed to be a cause of confusion before as it could be mistaken to mean that this file should always contain an empty object. Maybe we could simply write:

> Then you can have separate `options.json` and `options-test.json` files and load them as follows:

We should also clarify in the original part where it says to create the `options.json` file containing an empty object that this is only temporary to show how to load the file and that options can be added later to that file.

book/src/components/resulttable.md
```diff
@@ -50,8 +50,34 @@ To use the table, define the `tableOptions` and include the component in your HT
...
+    },
+    {
+      title: "Patients",
+      dataKey: "patients",
+    },
+  ],
+};
+
+options.tableOptions = tableOptions;
+```
```
timakro: Same point as before, I wouldn't duplicate the options example in TypeScript.

timakro (changes requested):

book/src/components/resultsummary.md
mmwaikar: But anyone who is using Lens for the first time won't realize in which file to include the JSON which we show in this particular page. Hence I thought of linking to the `Options and catalogue` section. If we don't do that then we'll have to show the entire code of how and where to define the options file and how to load the options from the etc.

By linking to this section, it becomes apparent that this (current) step is in continuation to what was followed in an earlier section or exercise.

book/src/components/resultsummary.md
mmwaikar: OK, I'll delete this - I just kept it because this thing worked for me.

book/src/components/resulttable.md
mmwaikar: OK, I'll delete this - I just kept it because this thing worked for me.

book/src/guide/new-app.md
mmwaikar: Mostly agree. I think it is obvious that the initially empty file will be populated later, but it was not clear in which part? 😊

book/src/components/resultsummary.md
timakro: I like the link I would just remove the `src/config/options.json` filename. Just like you did in `resulttable.md`.

book/src/guide/new-app.md
timakro: I think we should find some way to introduce the term "Lens options" here unambiguously, so that whenever some part of the docs instructs you to "add something to the Lens options" (very common) you know what to do.

book/src/components/resultsummary.md
mmwaikar: Done - deleted the section.

book/src/components/resulttable.md
mmwaikar: Done - deleted the section.

book/src/components/resultsummary.md
mmwaikar: Done.

book/src/guide/new-app.md
mmwaikar: Done - added a NOTE section.

book/src/guide/new-app.md
```diff
@@ -83,6 +83,8 @@ Now run `npm run dev` and open <http://localhost:5173/> in your browser. You sho
 
 Your application must pass two objects to Lens. The [LensOptions](https://samply.github.io/lens/docs/types/LensOptions.html) object contains general configuration options and the [Catalogue](https://samply.github.io/lens/docs/types/Catalogue.html) object describes what users can search for. You can define these objects in TypeScript but many applications in the Samply organization define them in JSON files.
 
+> **NOTE:** Whenever we talk about configuring Lens through a JSON file, we are talking about the below mentioned `options.json` file. This file is being initialized with an empty object for now, but will be populated in the later parts of this book.
```
timakro:
```suggestion
**NOTE:** The `options.json` file is being initialized with an empty object for now, but will be populated in the later parts of this book. Whenever the book tells you to add something to the Lens options it is implied that you add it to this file.
```

Please also move this below the catalogue example so it comes after the instructions to create `options.json`.

timakro (changes requested):

book/src/guide/new-app.md
mmwaikar: Done.

timakro:
Thanks!

PR #588, opened 2025-08-12
Fix reactivity

timakro:
Added siteResults as first parameter so the caller has to read the store. This is needed so the read is detected by svelte and the reader is automatically subscribed.

### DIFF
--- a/dev.svelte
+++ b/dev.svelte
@@ -7,6 +7,7 @@
         setSiteResult,
         getAst,
         resetDiagrams,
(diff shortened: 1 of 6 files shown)


patrick-skowronek (approved):
Good Work, but this must be a oversight in svelte right?

timakro:
I think it comes down to the use of `get()` vs `$` operator to read a store. `get` does not track reactivity while `$` does, I think this is intended.

PR #617, opened 2025-09-02
chore: don't set query param when query is empty

### LINKED ISSUE #600: Only show query in the URL after search button is clicked

timakro:
Closes #600


patrick-skowronek (commented):
Code looks good. I'm not quite sure if this is the best handling of this issue. Should we update the url after a change or when the query is triggered?

timakro:
I prefer it like this updating when changed (as opposed to only when search button is clicked)

patrick-skowronek (approved):

Issue #612, opened 2025-08-27
Track failed sites


timakro:
Spinner now fixed as https://github.com/samply/lens/pull/618 is merged. Converting this issue to track the "failed sites" use case. Original description below.

---

I think the spinner only works in the `dev.svelte` demo, not in a real project, because in a real project `loading` is set to `false` immediately as there are no claimed sites (yet) when the search button is clicked:

https://github.com/samply/lens/blob/8748e90b8f032cd6e904cd3d17ccf591939f1004/src/components/informational/QuerySpinner.wc.svelte#L14-L26

patrick-skowronek:
It worked somewhat for my own Project. I would argue that a project that doesn't use the spot or blaze interface should set the site as claimed after the search-button is triggered. That would also imply that they set it to permfail when the request fails.

We could also change the logic that the spinner listen to lens-search-button triggers and stop when something is triggered, like lens-search-finished. 

I would like to hear your view on this.

timakro:
I think we should derive it like this:

```ts
let loading = $derived.by(() => Array.from($siteStatus.values()).any(status => status === "claimed"));
```

This would mean that the spinner only starts spinning after the first site has been claimed but I think that's fine.

We should probably also bring back failure status via `markSiteFailed()` or similar to prevent failing sites from staying claimed forever and keeping the spinner spinning. I know I didn't want to add failure status is the past but I think we should now that we have a use for it. I think the table should also display "failed" when a site has been marked as failed.

patrick-skowronek:
I don’t fully remember why we were against this. From what I recall, when we query Spot, Beam would return either a permfail or tmpfail, which was enough for our applications. I still think we shouldn’t display fail in the table, since that would likely confuse users. It also feels confusing if a site briefly appears in the table and then disappears.

My first instinct would be to add a spinner in the table to indicate that results are still coming in, while showing the sites/data sources that have already responded.

That said, here are my remarks—we should define a proper failure state and clarify how the UI should behave when it happens:

1. Table – Do we show the site (with a failure indicator), or omit it entirely?
2. Result summary – Do we still count a failed site in the total site count?
3. Graphs – How do we handle failures, there?
4. Spinner – Should handle fails as well?
5. Non-LensResult values – How do we handle cases where Spot or a user wants to set something else?
6. Error handling – When do we display an error toast? For example: if all sites/data sources fail, should we show a single error notification?

timakro:
[...]

PR #618, opened 2025-09-02
chore: show spinner when there are claimed sites

timakro:
This gets rid of any raciness that would lead to the spinner not showing in some circumstances. The question is if this is the logic we want. I would argue it is a pragmatic solution and good enough. Some flaws with this:
* Spinner is not showing until first site is claimed
* If there are no claimed sites temporarily and then more results come in the spinner will stop spinning temporarily: e.g. berlin claimed -> berlin succeeded -> hamburg claimed -> ... (We can't really solve this without knowing which sites will be queried in advance.)

Related discussion: #612

### DIFF
--- a/src/components/informational/QuerySpinner.wc.svelte
+++ b/src/components/informational/QuerySpinner.wc.svelte
@@ -9,21 +9,9 @@
 
     let { size = "20px" } = $props();
 
-    let loading = $state(false);
-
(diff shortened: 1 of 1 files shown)


patrick-skowronek (approved):
Code looks good! I think this is the way to go. If one of the data sources are claimed it should inform the user

Issue #304, opened 2025-04-01
[Feature]: Query explain button should be more user friendly


patrick-skowronek:
### General Summary

The query explain button(lens-info-button with showquery=true) it is very suited for users who understand first order logic. We should improve the output that a clinical researcher without the deeper knowledge of first order logic understand the query.

### Is this Breaking?

No

MatsJohansen87:
How about something like this?

<img width="474" height="410" alt="Image" src="https://github.com/user-attachments/assets/aa48c579-2093-4b21-b3a4-bf851cf9cfc7" />

patrick-skowronek:
Hey @MatsJohansen87 , that looks neat. I have some input, that might improve it. 

- Use Group 1 and 2 instead of A and B, otherwise the user will search for group A
- The title could be better, I would maybe switch match with find? Find data with these attributes:
- I think in a group we can visualize the all must apply, maybe a and in front of the list?
- What does the ANY of mean? to be concise should it be or list? Or if there are more then 2 we say any with a comma separated list?

Some thoughts

MatsJohansen87:
"I think in a group we can visualize the all must apply, maybe a and in front of the list?"
wouldn't that be a step back to first order logic?

how about "Find samples matching one or more of the following groups:" ?

MatsJohansen87:
And how about entities? Should we just display the name of the entity or resolve it in a subgroup? We could also take the subgroup version for the chips and the other one for the general info, which would be a piece of work, but maybe worth to do.

<img width="474" height="410" alt="Image" src="https://github.com/user-attachments/assets/34e528f6-3bba-4585-a6b1-40cc9968ad8d" />

<img width="474" height="410" alt="Image" src="https://github.com/user-attachments/assets/1d7a61c7-fb60-4f4e-939b-9252acc1c0fc" />

timakro:
The first mockup seems most clear to me I specifically like the all-caps ALL and ANY words. Makes it very clear.

patrick-skowronek:
closed by #625

PR #625, opened 2025-09-03
Feature/rework query explain buttons

MatsJohansen87:
### Description

The query explain button(lens-info-button with showquery=true) it is very suited for users who understand first order logic. We should improve the output that a clinical researcher without the deeper knowledge of first order logic understand the query.

the whole query is displayed as discussed in the issue. the entities are displayed in a similar way.

Screenshot 2025-09-03 at 11 22 33

Screenshot 2025-09-03 at 11 21 56

### Related Issue

https://github.com/samply/lens/issues/304

---

### How Has This Been Tested?

UI user test

### Breaking

Maybe break negotiate formatting. Should not break functionality. Please test.


timakro:
I like that the aggregated value logic is only shown in the chip info button and in the query info button it just appears as "Gliom Grad 1". I think this breaks it up nicely and ensures that the query info button does not show overwhelming amount of info.

Regarding implementation I see your approach is very flexible (`useFullAggregatedValue` option and such) but I honestly would prefer a less-code approach and no additional data structures. Essentially directly render out the `QueryItem[][]` in a nested `{#each}` loop.

MatsJohansen87:
I added this flexibility also in regards to the query sent to the negotiation platform(s).
There maybe it could be useful to have both options. It's a tradeoff of some more code vs. flexibility for negotiation. also if someone wants to have the output without button(like in a custom component), they could do so.
I like things flexible, but that's often a "me-thing".
@patrickskowronekdkfz your thoughts?

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -28,33 +30,57 @@
...
 {:else}
     <div part="lens-query-explain-button">
-        <InfoButtonComponent
-            {message}
-            buttonSize="25px"
-            alignDialogue="left"
-            dialogueMaxWidth="600px"
-        ></InfoButtonComponent>
+        <InfoButtonComponent buttonSize="25px" alignDialogue="left">
+            {#if humanreadableQueryObject && humanreadableQueryObject?.groups.length > 0}
```
patrick-skowronek: Something with the logic is wrong. The else case is not working in the demo.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -69,4 +95,32 @@
...
+        list-style: none;
+        padding: 0;
+        margin: 0;
+        text-align-last: left;
+    }
+    [part~="lens-query-explain-group-item"] {
+        margin-bottom: var(--gap-xs);
+    }
+
+    [part~="lens-query-explain-bottom-level-items"] {
```
patrick-skowronek, line 150: This should do a css breakline 
Bildschirmfoto 2025-09-03 um 14 43 01
It could overflow the box :)

src/stores/datarequests.ts
```diff
@@ -13,85 +17,212 @@ export function getSelectedSites(): string[] {
...
-        queryLayer === null ||
-        !("children" in queryLayer) ||
-        ("children" in queryLayer &&
-            (queryLayer.children === null ||
-                queryLayer.children.length === 0 ||
-                queryLayer.children[0] === null))
-    ) {
-        return humanReadableQuery;
-    }
+export function getHumanReadableQuery(options?: {
```
patrick-skowronek: I would split this in a pure string function, as needed for the data request things, and an object function. I feels a bit over engineered

patrick-skowronek (changes requested):
I think we ware going into the right direction. I think the code is might be a bit over engineered and could be simplified. Maybe we don't need the additional data structures. I would suggest we make a meeting and look into the code.

patrick-skowronek:
I'm not quite sure if you are done with your refactoring. But this is a issue I found
Bildschirmfoto 2025-09-05 um 09 44 39

MatsJohansen87:
yes, i'm still on it

patrick-skowronek (approved):
The Code looks good, I would change the Docs after this PR.

src/components/buttons/InfoButtonComponent.wc.svelte
```diff
@@ -118,6 +116,8 @@
         padding: var(--gap-xs);
         border: solid 1px var(--blue);
         border-radius: var(--border-radius-small);
+        text-align-last: left;
```
timakro: This should be `text-align` if there is no specific reason to use `text-align-last`.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -27,34 +24,91 @@
...
+    </InfoButtonComponent>
 {:else}
     <div part="lens-query-explain-button">
         <InfoButtonComponent
-            {message}
             buttonSize="25px"
             alignDialogue="left"
-            dialogueMaxWidth="600px"
-        ></InfoButtonComponent>
+            {inSearchBar}
```
timakro: This is always `false` so you can remove it and also remove it in the `Props` interface at the start of the file.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -27,34 +24,91 @@
...
+                            <div
+                                part="lens-query-explain-multi-row-message-group-item"
+                            >
+                                {getCategoryFromKey($catalogue, valueItem.value)
+                                    ?.name
+                                    ? getCategoryFromKey(
+                                          $catalogue,
+                                          valueItem.value,
+                                      )?.name
+                                    : valueItem.value}: {valueItem.name}
```
timakro: Use nullish coalescing operator `??` then you don't have to repeat the function call.

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -610,11 +612,15 @@
     }
 
     [part~="lens-searchbar-chip-item"] {
-        display: inline-flex;
+        display: grid;
+        grid-template-columns: 1fr auto auto;
+        grid-gap: var(--gap-xxs);
```
timakro: Is there a specific reason to use grid layout here? For one dimensional layouts flexbox seems more appropriate.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -69,4 +123,58 @@
         border: solid 1px var(--light-blue);
         border-radius: var(--border-radius-small);
     }
+
+    [part~="lens-query-explain-single-row-message"] {
+        text-align: left;
+        line-break: anywhere;
```
timakro: Not sure about `line-break: anywhere`. I think we should avoid this unless we expect extremely long words without spaces.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -69,4 +123,58 @@
...
+    }
+
+    [part~="lens-query-explain-bottom-level-items"] {
+        padding-left: var(--gap-xs);
+        list-style: none;
+    }
+
+    [part~="lens-query-explain-bottom-level-item"] {
+        margin-bottom: var(--gap-xxs);
+        line-break: anywhere;
```
timakro: Same here.

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -610,11 +612,15 @@
...
         align-items: center;
-        gap: var(--gap-xxs);
     }
 
+    [part~="lens-searchbar-chip-item-text"] {
+        line-break: anywhere;
```
timakro: Same here.

src/helpers/translations.ts
```diff
@@ -2,7 +2,11 @@ import { get } from "svelte/store";
 import type { Texts } from "../types/options";
 import { lensOptions } from "../stores/options";
 
-/** Translate a key into a human readable text. */
+/**
+ * Translate a key into a human readable text depending on the language set in options
+ * @param key the key to translate
+ * @returns the human readable text
```
timakro, line 7: This doesn't really add any information. I think it is okay to omit `@param` and `@returns` in cases where they don't help understanding.

src/stores/datarequests.ts
```diff
@@ -95,3 +101,160 @@ export const getHumanReadableQuery = (): string => {
...
+interface GetHumanReadableQueryAsFormattedString {
+    printAggregatedValues?: boolean;
+}
+
+/**
+ * Formats the query into a human-readable string.
+ * @param param0 options.printAggregatedValues if true, the deep values of entities will be parsed and shown
+ * @returns a formatted string representation of the query
+ */
+export function getHumanReadableQueryAsFormattedString({
```
timakro: This is no longer used and can be deleted.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -27,34 +24,91 @@
...
-    ></InfoButtonComponent>
+{#if queryItem}
+    <InfoButtonComponent buttonSize="18px" inSearchBar={true}>
+        {#if typeof queryItem?.values[0].value === "string"}
+            <div part="lens-query-explain-single-row-message">
+                {queryItem.name}: {queryItem.values[0].value}
+            </div>
+        {:else if "min" in queryItem.values[0].value || "max" in queryItem.values[0].value}
+            <div part="lens-query-explain-single-row-message">
+                {queryItem.name}: {getMinMax(queryItem.values[0].value)}
```
timakro: You can use `queryItem.name` it already contains the formatted string.

timakro (changes requested):

src/components/buttons/QueryExplainButtonComponent.wc.svelte
MatsJohansen87: I would leave it in. Even if the user inputs something stupid long, that doesn't really make sense (like in the first name field), the ui shouldn't break. maybe then a better way would be to limit the input field?

timakro: It's just that you get mid-word line breaks even if not necessary. If there was something that would try to break at word boundaries if possible and only do mid-work breaks if unavoidable I wouldn't be so against it.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: you're right, but adding a gap was needed. changed it back and added a flex gap.

src/stores/datarequests.ts
MatsJohansen87: this is, as discussed, the function to replace the old first order logic output (which is still used in bbmri) in the future. it's currently not used in the ui. maybe we should add a button to show it?

src/components/buttons/QueryExplainButtonComponent.wc.svelte
MatsJohansen87: like this?

Screenshot 2025-09-11 at 12 08 18

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: After changing back and messing with the long names i saw again what broke. The buttons shrink. This had an easy fix with grid.

Screenshot 2025-09-11 at 12 22 31

timakro: Yeah flexbox does even shrink elements with fixed `width` which can be annoying.

```css
.container-name {
  InfoButtonComponent {
    flex-shrink: 0;
  }
}
```

Or similar should fix it.

src/stores/datarequests.ts
timakro: I'd say remove it if not used. Who knows when/if we'll need it.

MatsJohansen87: @patrickskowronekdkfz your thoughts?

patrick-skowronek: Please exposed the function

demo.svelte
```diff
@@ -422,7 +423,17 @@
...
-                <lens-negotiate-button></lens-negotiate-button>
+                <!-- <lens-negotiate-button title="Request Data"></lens-negotiate-button> -->
+                <button
+                    style="margin: 20px 0; padding: 10px 20px; border-radius: 5px; background: #0047b9; border: none"
+                >
+                    <a
+                        style="color:white; text-decoration: none; font-size: 16px"
+                        href="mailto:request@example.com?body={encodeURIComponent(
+                            getHumanReadableQueryAsFormattedString(),
+                        )}">Request Data</a
```
timakro: This doesn't update. I think the best solution would be to dynamically create a `mailto:` anchor element `onclick` and programmatically click it.

src/stores/datarequests.ts
```diff
@@ -95,3 +101,141 @@ export const getHumanReadableQuery = (): string => {
 
     return humanReadableQuery;
 };
+
+interface GetHumanReadableQueryAsFormattedString {
+    printAggregatedValues?: boolean;
+}
```
timakro: No need for interface, can just write a normal function.

src/stores/datarequests.ts
```diff
@@ -12,8 +14,12 @@ export function getSelectedSites(): string[] {
     return get(datarequestsStore);
 }
 
+const catalogueForNameExtraction = get(catalogue);
```
timakro: Did you test this? I would assume this reads an empty catalogue because it runs before the app has set one. I'd suggest to remove the variable and just call `get(catalogue)` when you need it.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -5,56 +5,99 @@
...
 </script>
 
-{#if inSearchBar}
-    <InfoButtonComponent {message} buttonSize="18px" inSearchBar={true}
-    ></InfoButtonComponent>
+{#if queryItem}
+    <InfoButtonComponent buttonSize="18px" inSearchBar={true}>
+        {#if typeof queryItem?.values[0].value === "string"}
+            <div part="lens-query-explain-single-row-message">
+                {queryItem.name}: {queryItem.values[0].value}
```
timakro, line 29: This only shows the first value if there are more than one.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -5,56 +5,99 @@
...
+            </div>
+        {:else if "min" in queryItem.values[0].value || "max" in queryItem.values[0].value}
+            <div part="lens-query-explain-single-row-message">
+                {queryItem.name}: {getMinMax(queryItem.values[0].value)}
```
timakro: This only shows the first value if there are more than one.

Also you can use `queryItem.values[0].name` instead of building the string here again.

timakro (changes requested):

src/components/catalogue/NumberInputComponent.svelte
```diff
@@ -56,7 +42,10 @@
                 type: element.type,
                 values: [
                     {
-                        name: buildName(),
+                        name: getMinMax({
+                            min: from as number,
+                            max: to as number,
```
timakro: You are casting the possible `null` away here which is bad. The `getMinMax()` would have to handle `null` I suppose. But generally I would tend to not pull out a function as trivial as this into its own file anyways.

demo.svelte
patrick-skowronek: I somewhat agree. I would rather use the following flow. Use the LensRequestButton. have a function that generates a mailto and use a click event to trigger it.

MatsJohansen87: Yes, seems the way to go.

src/stores/datarequests.ts
MatsJohansen87: right, legacy from the object || string function

src/stores/datarequests.ts
MatsJohansen87: Works fine

Screenshot 2025-09-15 at 12 11 34

timakro: That variable is only used by your string generating functions though, not by the current UI.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
MatsJohansen87: The array allways has only one entry when it is built as query object. Looking through old code (lens-angular) this was the structure we overtook from there to be conform with ast and cql converters. After that we removed the cql translators from lens and refactored the query to ast translator, so this is a legacy thing and should be refactored. That would include a lot of files but shouldn't be breaking.

demo.svelte
```diff
@@ -383,6 +384,17 @@
         localStorage.setItem("language", lang);
         window.location.reload();
     }
+
+    const sendQueryAsEmail = () => {
```
patrick-skowronek:
```suggestion
    window.addEventListener("lens-negotiate-triggered", () =>
```

demo.svelte
```diff
@@ -422,7 +434,13 @@
             </div>
             <div id="result-table" class="card">
                 <lens-result-table pageSizeSwitcher={true}></lens-result-table>
-                <lens-negotiate-button></lens-negotiate-button>
```
patrick-skowronek: I would use the lens-negotiate-button, then we test it and show how to use it. And we don't the styling.

src/components/buttons/QueryExplainButtonComponent.wc.svelte
```diff
@@ -5,56 +5,98 @@
...
+    import { translate } from "../../helpers/translations";
+    import { catalogue, getCategoryFromKey } from "../../stores/catalogue";
     import { queryStore } from "../../stores/query";
+    import type { QueryItem } from "../../types/queryData";
     import InfoButtonComponent from "./InfoButtonComponent.wc.svelte";
 
     interface Props {
         noQueryMessage?: string;
         queryItem?: QueryItem | undefined;
         /** Query explain button in the search bar is smaller, white, and has no border */
```
patrick-skowronek: Remove comment

patrick-skowronek (changes requested):
I think it is almost ready. Just some small things

demo.svelte
```diff
@@ -422,7 +434,8 @@
...
+                <lens-negotiate-button title="Request Data"
+                ></lens-negotiate-button>
```
timakro:
```suggestion
                <lens-negotiate-button></lens-negotiate-button>
```

demo.svelte
```diff
@@ -527,6 +540,12 @@
         }
     }
 
+    #result-table {
+        display: flex;
+        flex-direction: column;
+        gap: 20px;
```
timakro:
```suggestion
        gap: var(--gap-s);
```

timakro (changes requested):

patrick-skowronek (approved):

timakro (approved):

PR #631, opened 2025-09-04
chore: make pagination button stay in position on last result page

### LINKED ISSUE #573: Table page forward/backward button should not move

timakro:
This PR fills up the result table with invisible rows so the height stays fixed. This has several advantages:
* The pagination button stays in place
* The layout does not shift when the table initially fills up with results

I think there wasn't a good reason to have the rows as a separate component so I merged them into the table which made this change easier.

https://github.com/user-attachments/assets/e20a04fa-f7f4-4d3e-96c5-ff572a624012

Closes https://github.com/samply/lens/issues/573 (except I didn't add sort arrows for the other columns)


patrick-skowronek (changes requested):
The code looks good. I have one styling issue. I would always display the lines between each row. Then it would look more consistent.

timakro:
> I would always display the lines between each row.

I thought this would look terrible but honestly upon trying it I like it.

MatsJohansen87:
If a user selects 50 sites per page it just looks awful imho.

Maybe this can be solved by placing pagination on top, so it doesn't jump while browsing through the pages.

Screenshot 2025-09-05 at 11 40 42

Screenshot 2025-09-05 at 11 21 36

timakro:
Yeah we could try how that looks. Also why is there no pagination on your screenshot? For me it always shows even if there are no results yet.

timakro:
One option we could think about is having a fixed height mode with a page size useful for grid layouts (think CCP explorer) and a variable height mode without paging that just shows all the rows useful for non-grid layouts (think BBMRI sample locator). That could involve removing the page size switcher whose usefulness I would question anyways since we always know pretty accurately how many sites there are in a given explorer and can tweak the layout accordingly.

MatsJohansen87:
> Yeah we could try how that looks. Also why is there no pagination on your screenshot? For me it always shows even if there are no results yet.

It was off screen, because the 50 lines were so long

> One option we could think about is having a fixed height mode with a page size useful for grid layouts (think CCP explorer) and a variable height mode without paging that just shows all the rows useful for non-grid layouts (think BBMRI sample locator). That could involve removing the page size switcher whose usefulness I would question anyways since we always know pretty accurately how many sites there are in a given explorer and can tweak the layout accordingly.

i think that's a valid point. the project will configure it anyways in the options, so no switcher needed. 

Other ideas:

- Set the switcher to 5 | 10 | all and give the project a config to choose default.

- Give the project the possibility to set the switcher sizes. This could reduce empty lines to a minimum
   e.g.  with 13 sites 5 | 10 | all ->  Pages: 5|5|3 or 10|3 or 13 
   e.g. with 42 sites: 7 | 14 | all -> Pages: 7|7|7|7|7|7 or |14|14|14 or 42 (though maybe the uneven values could look messy for the user)

patrick-skowronek:
After thinking about this, if the project wants to set it to 50 but only has 20 sites, this is a project issue, not a library issue. If you want to display the first 50 hits, I would build a layout that can accommodate this. The library behaves as expected. I would not start with some kind of prop that lets the user decide where to put the paginator.

MatsJohansen87:
currently it's hardcoded by us. the user still has the option to display 50 sites regardless of what the project has chosen as page size.
```
    const pageSizeOptions: number[] = Array.from(
        new Set([10, 25, 50, pageSize]),
    ).sort((a, b) => a - b);
    let currentPageSize = $state(pageSize);
```

PR #633, opened 2025-09-04
Prevent layout shift from query modified and add default messages

timakro:
I think preventing the layout shift is a good idea but of course it also wastes space. What do you think?


patrick-skowronek (approved):
Good catch. Maybe we should add it to the doc of the component?

timakro:
Good point, docs updated.

src/components/informational/ModifiedSearchComponent.wc.svelte
```diff
@@ -5,14 +5,18 @@
...
-{#if $queryModified}
-    <div part="lens-query-modified-display-wrapper">
-        <slot></slot>
-    </div>
-{/if}
+<div
+    part="lens-query-modified-display-wrapper"
+    style="visibility: {$queryModified ? 'visible' : 'hidden'}"
+>
+    <slot>
```
MatsJohansen87, line 16: do we still need the slot here, when the text comes from translations?

timakro: It allows apps to override the text if they want.

timakro: On second thought, they could also override via the translation API though..

MatsJohansen87: i would give the user only one way, to avoid confusion

patrick-skowronek: I would also vote for removal. They also can style the box, so why put there something in?

PR #656, opened 2025-09-11
New README

timakro:
I wanted to make the README less noisy/AI-ish for a while. What do you think?


timakro:
Also maybe we could have the README as the first page of the book?

MatsJohansen87:
> Also maybe we could have the README as the first page of the book?

I think that's a good idea.

MatsJohansen87:
Maybe we should mention somewhere in the readme, that it's a framework agnostic web component library.

README.md
```diff
@@ -1,111 +1,35 @@
-# samply.lens
+# Lens
 
-**samply.lens** is a front-end library that provides essential functionalities for building search, exploration, and visualization web applications. It offers a set of flexible building blocks while remaining open to user-specific extensions.
+Lens is a front-end library for building federated search tools in the medical space to answer questions such as:
```
patrick-skowronek:
```suggestion
Lens is a front-end library for building search tools in the medical space to answer questions such as:
```

patrick-skowronek: Not all tools are federated 😉

patrick-skowronek (changes requested):

timakro:
> Maybe we should mention somewhere in the readme, that it's a framework agnostic web component library.

Not sure we should mention it as I'm not so attached to that idea. I think we should strongly push every one in our org that uses Lens to use Svelte.

patrick-skowronek (approved):

MatsJohansen87:
> Not sure we should mention it as I'm not so attached to that idea. I think we should strongly push every one in our org that uses Lens to use Svelte.

I agree that we should direct users to use svelte whenever possible. But a goal we set was to make a framework agnostic library, so projects outside of our organisation can use it in their preferred environment. Maybe they want to integrate it in an existing project as a small part on a sub site. We could suggest the usage of svelte, as the template/demo is also is written in svelte but mention, that it can be used elsewhere. 
Maybe it suffices to mention it as "web component library". This indicates by itself that it's agnostic.

timakro:
Yeah I mean it doesn't mention Svelte anywhere right now and just says "front-end library". Shouldn't deter anyone from trying to use it in a vanilla JS site. We can say "web component library" instead if you want.

Issue #662, opened 2025-09-16
[Refactor]: queryItem


MatsJohansen87:
### General Summary

The QueryItem given to the QueryExplainButtonComponent from the SearchbarComponent is built newly with only one item in the values array. This is confusing and should be refactored to give only name and value to the button, so the values[0].value isn't needed anymore.

### Is this Breaking?

No

PR #664, opened 2025-09-16
Refactor/query object

MatsJohansen87:
### Description
The object structure in the QueryExplainButtonComponent was messy and confusing.

Replaces the QueryObject with singluar values to make code more understandable
```
queryItemName?: string | undefined;
queryItemValue?: QueryValue | undefined;
```

Additional fix: getMinMax() was behaving strangely with strings and zeroes, because of a refactor to handle null values. now truthy and falsy statements are used to display the values correctly

### Related Issue

https://github.com/samply/lens/issues/662
---

### How Has This Been Tested?
UI user test


patrick-skowronek (changes requested):
Should we give getMinMax a better function name? 
The change should be documented in the docs and is then blocked by #665

timakro:
Won't the falsy check handle `0` incorrectly?

MatsJohansen87:
Not in this case. We want to have falsy values to make `<=` or `>=`. This includes an empty input as well as a `0`. If a user inputs `between 0 and 15` we want to have displayed `<=15`

MatsJohansen87:
> Should we give getMinMax a better function name? The change should be documented in the docs and is then blocked by #665

`getMinMaxValueString()`?

patrick-skowronek:
> > Should we give getMinMax a better function name? The change should be documented in the docs and is then blocked by #665
> 
> `getMinMaxValueString()`?

`getQueryValueMinMaxString()` would be my proposal

timakro:
> Not in this case. We want to have falsy values to make `<=` or `>=`. This includes an empty input as well as a `0`. If a user inputs `between 0 and 15` we want to have displayed `<=15`

In the AST these are different so they should be distinguishable in the UI:
image
Negative numbers are supported (although not used I think) so the way it currently works is good in my opinion.

timakro:
Honestly I think we should move the min-max formatting function back to the components. "Duplicating" six lines is better than having an extra file with only a six line function.

patrick-skowronek:
> Honestly I think we should move the min-max formatting function back to the components. "Duplicating" six lines is better than having an extra file with only a six line function.

I think we can move this function somewhere else, I agree having one file is a bit much for it. The function is used at more points in the code

patrick-skowronek (approved):

PR #669, opened 2025-09-19
Failed sites

### LINKED ISSUE #612: Track failed sites

timakro:
Closes #612

### DIFF
--- a/book/src/guide/results.md
+++ b/book/src/guide/results.md
@@ -30,7 +30,7 @@ The `totals` field contains the total number of patients, samples, etc. The `str
 
 ## Passing results to Lens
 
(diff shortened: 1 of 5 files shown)


patrick-skowronek (commented):
Thx @timakro for the PR. I like the implementation, but I would like to discuss the logic. If a site fails for a reason -> query fails, network issue and so on; it would just disappear. Would it not make sense to provide a function that sets the site to fail and informs the user? I can understand that certain applications would hide the site, but there would be use cases where it fails for a reason and I think we should think about them. What are your thoughts?

timakro:
I think removing the site from the result table and result summary makes sense. Apps could show an error toast in addition. But if you think a totally different behavior would be better I'm open to ideas.

patrick-skowronek (approved):

PR #671, opened 2025-09-21
feat: add function to set the facetcount

### LINKED ISSUE #636: [Feature]: Make facetcounts set via function
### General Summary

The facetcounts are a nice feature in lens. I noticed in the dev environment, it is set via an internal access to the store directly. Outside of the dev environment it is only possible to set it via spot. As with the backends we should have a function to get the store and the project should set the store. We still could provide the function to get it from spot, but set it in the project via `setFacetCounts()`.

### Is this Breaking?

Yes

patrick-skowronek:
### Description

closes #636

### DIFF
--- a/src/types/facetCounts.ts
+++ b/src/types/facetCounts.ts
@@ -1,1 +1,0 @@
-export type FacetCounts = Record<string, Record<string, number>>;
--- a/src/types/facetCounts.ts
+++ b/src/types/facetCounts.ts
@@ -1,0 +1,1 @@
+export type FacetCounts = Record<string, Record<string, number>>;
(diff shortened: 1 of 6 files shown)


timakro (approved):

src/types/facetCounts.ts
```diff
@@ -0,0 +1 @@
+export type FacetCounts = Record<string, Record<string, number>>;
```
timakro: Move this to `stores/facetCounts.ts`. The reason why we had types in the `types` directory previously was technical so we could bundle all the types into a `.d.ts` but now we don't have to do that anymore so I think we should keep the types with the code that uses them.

It would also be nice if the type had a doc string with an example.

timakro (changes requested):

src/types/facetCounts.ts
patrick-skowronek: @timakro  how about we move then all the types to the stores?

timakro: I think I already moved some as part of some other PRs when I worked on something related. Big files like `types/options.ts` can stay as separate files, especially if JSON schema is generated from them. But especially small files with only a few lines I think we should move.

timakro (approved):

Issue #498, opened 2025-07-10
[Feature]: Allow number/date/string input in search bar


patrick-skowronek:
### General Summary

We should add support for entering numbers, dates, and string-based values directly into the search bar.
For example, if a user wants to search for body weight, the only current options are navigating through the catalogue or using a specific code.

Ideally, if a user types "body weight" into the search bar, it should appear as a selectable option — even for structured fields like numeric values.

### Is this Breaking?

No

PR #689, opened 2025-09-26
Feautre/add input fields to searchbar

MatsJohansen87:
### Description
Add input fields to Search bar
### Related Issue
https://github.com/samply/lens/issues/498

---

### How Has This Been Tested?
UI user test


src/components/buttons/SearchButtonComponent.wc.svelte
```diff
@@ -11,13 +11,21 @@
     interface Props {
         title?: string;
         disabled?: boolean;
+        resetSearchBarTextInputs?: boolean;
```
timakro: I think this doesn't have to be configurable and could be the standard behavior.

src/components/catalogue/AddButton.svelte
```diff
@@ -1,8 +1,18 @@
 <script lang="ts">
-    let { ...props } = $props();
+    interface Props {
+        inSearchBar: boolean;
+        [key: string]: unknown;
```
timakro: How hard would it be to get rid of this "anything else" prop? I kind of regret adding it because Svelte always complains about it.

src/components/catalogue/DatePickerComponent.svelte
```diff
@@ -57,11 +72,55 @@
...
+    }
+
+    let form: HTMLElement;
 </script>
 
-<form part="lens-date-input-form" {onsubmit}>
+<form
+    part="lens-date-input-form"
+    bind:this={form}
+    {onsubmit}
```
timakro: Originally this used the behavior that if there is a single `<button>` in a `<form>` then this button is considered the submit button. I like that it is more explicit now. You can remove the event here and rename that function so it doesn't look like an event handler anymore and also remove the parameter of type `SubmitEvent`.

src/components/catalogue/DatePickerComponent.svelte
timakro, line 1: I think the DatePickerComponent should be less concerned with the search bar functionality.

To keep up/down arrow navigation functionality I think it would be fine to use a capture phase event handler (https://svelte.dev/tutorial/svelte/capturing) in the search bar drop down to intercept up and down arrows before they even reach the text input using `event.stopPropagation()`. For numbers this would be fine in my opinion. I don't consider the up/down arrow to increment/decrement numbers important. I hope the date picker calendar arrow key navigation would not break.

I think you could use a similar strategy for detecting focus without touching the DatePickerComponent. Or alternatively just add a `onfocus` prop to the DatePickerComponent and handle everything else outside of the component.

src/components/buttons/StoreDeleteButtonComponent.svelte
```diff
@@ -14,9 +14,10 @@
             index: number;
             item?: QueryItem;
         };
+        onFocusOutOfSearchBar?: (event: FocusEvent) => void;
     }
 
-    let { itemToDelete }: Props = $props();
+    let { itemToDelete, onFocusOutOfSearchBar = () => {} }: Props = $props();
```
timakro: This prop is unused.

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -286,8 +315,52 @@
                 extractTargetGroupFromInputValue(),
             );
         }
+        if (event.key === "Tab" && !event.shiftKey) {
+            if (activeDomElement) {
+                event.preventDefault();
+                focusInSearchBarOption(0, focusedItemIndex, activeDomElement);
+            }
+        }
```
timakro: Is focusing the first number/date/string input? Wouldn't that be the next tab target anyways?

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -400,13 +465,75 @@
                 window.history.replaceState({}, "", url.toString());
             }
         });
+
+        let clickInsideFlag = false;
+
+        searchBarContainer.addEventListener("mousedown", (event) => {
```
timakro: Before this was a `onmousedown` handler on the item which made a lot of sense. Now it is a handler on the entire search bar container. Can we move this back to the item?

Also, use Svelte-style `onevent` bindings instead of `addEventListener`.

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -474,91 +598,210 @@
...
+                                        focusInSearchBarOption(
+                                            elementIndex,
+                                            i,
+                                            optionElements[i],
+                                        )}
+                                    {resetToEmptySearchBar}
+                                    {focusSearchbar}
+                                />
+                            </li>
+                        {/if}
```
timakro: Is the only difference here the `-focused` class? If so only make that class conditional instead of duplicating the entire block.

src/index.ts
```diff
@@ -6,7 +6,7 @@ export { setCatalogue } from "./stores/catalogue";
 export { setOptions } from "./stores/options";
 export { translate } from "./helpers/translations";
 export { showToast } from "./stores/toasts";
-export { resetDiagrams } from "./helpers/reset";
+export { resetDiagrams, resetSearchBarTextInputs } from "./helpers/reset";
```
timakro: Applications probably don't need this functionality.  I would not export it.

eslint.config.js
```diff
@@ -7,7 +7,7 @@ import svelte from "eslint-plugin-svelte";
 import svelteConfig from "./svelte.config.js";
 
 export default defineConfig([
-    globalIgnores(["dist", "book/book", "docs/assets"]),
+    globalIgnores(["dist", "book", "docs/assets"]),
```
timakro, line 10: Generally I would want to lint the book markdown files. Can we fix the issues without disabling linting of those files completely?

src/types/queryData.ts
```diff
@@ -47,3 +38,23 @@ export type SendableQuery = {
     query: QueryItem[][];
     id: string;
 };
+
+export type AutoCompleteCriterionItem = {
+    fieldType: string;
```
timakro:
```suggestion
    fieldType: "criterion";
```

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -400,13 +465,75 @@
                 window.history.replaceState({}, "", url.toString());
             }
         });
+
+        let clickInsideFlag = false;
```
timakro: This flag and the 10 ms timeout look really hacky. Can we find a better way to do this?

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -456,16 +583,13 @@
...
-            autoCompleteOpen = false;
-            inputValue = "";
-        }}
     />
-    {#if autoCompleteOpen && inputValue.length > 2}
-        <ul part="lens-searchbar-autocomplete-options">
+    {#if autoCompleteOpen && inputValue.length}
+        <ul part="lens-searchbar-autocomplete-options" bind:this={optionsList}>
             {#if inputOptions?.length > 0}
                 <!-- eslint-disable-next-line svelte/require-each-key -->
```
timakro: eslint-disable comment no longer required

timakro (changes requested):

src/components/buttons/SearchButtonComponent.wc.svelte
MatsJohansen87: Fine with me, was just intended as a nice to have.

src/components/catalogue/AddButton.svelte
MatsJohansen87: We can pass `{onkeydown}` and `{onclick}` as optional handlers. I don't see other stuff needing to get handled right now

src/components/catalogue/DatePickerComponent.svelte
MatsJohansen87: Right. I would give the `AddButton` a `type="button". Otherwise a default submit event will be fired from the form.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: The problem here is, that if you tab into the first element, the options are built and rendered again. This leads to a loss of focus, because the element you tabbed into doesn't exist anymore. Hence we need to wait for the next render and then set focus again. Maybe i can find a more elegant way, if i can get the list to not rerender. could maybe also fix the duplicate if:else of the list items and some other issues if i succeed. might be a bigger one 😅

MatsJohansen87: If an element is selected via arrows, the tab should go there. Hence we need to preventDefault. but i could remove the extra function.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: The big difference was the binding to `activeDomElement` which i could refactor.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: I was able move it to the component again with the refactor. For global events we could use `<svelte:window />` but it doesn't support custom events. So we would be mixing syntax for the same things, which i'm not a big fan of.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: solved in the big refactor

eslint.config.js
MatsJohansen87: I agree. but `book/book` doesn't exist anywhere, so this was unnecessary, right?

timakro: When you install mdbook and build the book locally it generates the HTML files there.

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -286,8 +319,53 @@
...
+    function handleClickInside(): void {
+        inside = true;
+    }
+
+    function handleClickOutside(): void {
+        if (!inside) {
+            autoCompleteOpen = false;
+        }
+        inside = false;
+    }
```
timakro, line 408: Are these click handlers needed? I tried without them and didn't noticed a difference

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -286,8 +319,53 @@
...
+        }
+    }
+
+    // needed as function to be passed to children
+    function focusSearchbar(): void {
+        searchBarInput.focus();
+    }
+
+    let searchBarContainer: HTMLElement;
+    let optionsList: HTMLUListElement;
```
timakro: This is unused

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -513,43 +613,83 @@
...
                              which will close the options before the click is finshed -->
                         <li
-                            part="lens-searchbar-autocomplete-options-item"
-                            onmousedown={() => selectItemByClick(inputOption)}
+                            bind:this={optionElements[i]}
+                            onmousedown={() =>
+                                addInputValueToStore(
+                                    inputOption,
+                                    extractTargetGroupFromInputValue(),
+                                )}
```
timakro, line 689: Does this do anything? The NumberInputComponent adds the item to the query itself.

timakro (commented):
The code in the search bar component looks much better now. And great that you added regex escaping to the search highlighting.

Do you think we can move some of the search bar and focus handling out of the input components? Maybe you can try my suggestion to intercept the arrow keys before the input components receive them. But I haven't thought a lot about that so I understand if you decide to stick with the current approach where you press escape to back out of the input component and then use the arrow keys.

src/components/search-bar/SearchBarComponent.wc.svelte
MatsJohansen87: this prevents closing the options if the click goes to onto a number/date/string input, but doesn't land on an input field. So if the user is hovering and has twitchy fingers like me or just misclicks it will close. May be a corner case but i like having it in.

MatsJohansen87: Should have handled the click onto an input element of the options but didn't work because the `inside` check was missing in `handleFocusOut()`. Added it and now it works as intended: when the click is outside the form elements but still inside of the input element it won't close.

MatsJohansen87:
> Maybe you can try my suggestion to intercept the arrow keys before the input components receive them. But I haven't thought a lot about that so I understand if you decide to stick with the current approach where you press escape to back out of the input component and then use the arrow keys.

I'm always careful when it comes to altering browser native behaviour. the user will expect the form fields to work the same way throughout the application. if we make that change, the input will behave differently from the ones in the catalogue. also it's a behaviour known to users for as long as they use the internet 😅 

But maybe the convenience for the user is high enough to make that change. I guess this would come down to a short user test. I will recruit some non-IT-people.

MatsJohansen87:
> Do you think we can move some of the search bar and focus handling out of the input components?

I tried from the beginning and failed miserably 😄. 100% open to elegant solutions.

timakro:
> I'm always careful when it comes to altering browser native behaviour.

I agree. In this case I think it might be worth it to prioritize the arrow keys but I'm not dead set on it.

> 100% open to elegant solutions

I'd try to use a capture phase event handlers (https://svelte.dev/tutorial/svelte/capturing) on the input components to catch the `focusin` and `keydown` events. If this works you could have the event handling code in the search bar component and reuse the handler functions between the different input component types. This should also get rid of some callback props if it works.

patrick-skowronek:
@MatsJohansen87 could you check why this feature is not working on safari? On chrome it works fine

MatsJohansen87:
I think i found an elegant way to move the focus logic to the searchbar component itself. I also changed my mind about the idea to change the active element instead of increasing or decreasing the numbers in number and date inputs and think it's more intuitive like this. So i also implemented that.
Please have another look. I also found the bugs where the first chip wasn't displayed and pressing enter didn't work.

src/components/catalogue/NumberInputComponent.svelte
```diff
@@ -39,8 +48,12 @@
         return "";
     }
 
-    function onsubmit(event: SubmitEvent): void {
-        event.preventDefault();
+    function addItem(): void {
+        if (!formVlaid) {
```
timakro, line 52: Instead of computing validity yourself use the return value of `reportValidity()` here. This will include validation of built-in HTML attributes like `required`, `min`, `max` and the custom errors you set with `setCustomValidity`. Right now for example the `min` HTML attribute does not work and it is possible to add body weight rage -40 through -10 in the demo.

src/components/catalogue/StringInputComponent.svelte
```diff
@@ -1,29 +1,45 @@
...
     });
 
-    function onsubmit(event: Event) {
-        event.preventDefault();
+    let formValid = $derived(validateForm(value));
+
+    function validateForm(value: string | null): boolean {
+        input.setCustomValidity("");
+        if (!value) {
+            input.setCustomValidity(translate("cannot_be_empty"));
```
timakro, line 31: See my comment on NumberInputComponent. You could use `required` HTML attribute instead of custom validation.

src/components/catalogue/NumberInputComponent.svelte
```diff
@@ -8,28 +8,37 @@
 
     let {
         element,
+        inSearchBar = false,
+        resetToEmptySearchBar = () => {},
```
timakro, line 12: To decouple this from the search bar I think it would be nicer to expose an event handler `onItemAdded` here and then register the callback from outside:

```html
<NumberInputComponent onItemAdded={resetToEmptySearchBar}>
```

src/components/catalogue/NumberInputComponent.svelte
```diff
@@ -57,10 +70,39 @@
...
+        // so this can handle the focus via mouse click instead of using another event listener
+        if (event.target === toInput) return;
+
+        const relatedTargetOutside =
+            event.relatedTarget instanceof Node &&
+            !form.contains(event.relatedTarget);
+
+        if (relatedTargetOutside) {
+            fromInput.focus();
+        }
```
timakro, line 101: Instead of handling these in the input components could you try my suggestion of handling them completely from the search bar using capture phase event handlers (https://svelte.dev/tutorial/svelte/capturing)?

For the enter key you can expose `addItem` to the outside.

The arrow keys are already handled from the outside if you now also call `event.preventDefault()` during the capture phase the input won't see the arrow key event so no need to ignore it here.

I'm not 100% sure about the focus logic but it seems it is already partially handled outside. I think it would be fine if you drill down to the first input element with a smart query selector from the outside and just focus that.

src/components/buttons/SearchButtonComponent.wc.svelte
```diff
@@ -18,6 +18,7 @@
     const onclick = (): void => {
         queryModified.set(false);
         window.dispatchEvent(new CustomEvent("lens-search-triggered"));
+        window.dispatchEvent(new CustomEvent("reset-all-searchbar-inputs"));
```
timakro, line 21: Currently only NumberInputComponent seems to respect this. Do we even need this as when search button is pressed the search bar should loose focus and thus all inputs should be destroyed and therefore their contents reset?

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -257,16 +290,24 @@
         return index;
     };
 
+    let searchBarInputHasFoucs = $state(true);
```
timakro, line 290:
```suggestion
    let searchBarInputHasFocus = $state(true);
```

src/components/search-bar/SearchBarComponent.wc.svelte
```diff
@@ -286,8 +327,65 @@
...
     };
 
+    function handleFocusIn() {
+        autoCompleteOpen = true;
+        let focusedListItem = optionElements.find(
+            (element) => element && element.matches(":focus-within"),
+        );
+        if (focusedListItem) {
+            focusedItemIndex = optionElements.indexOf(focusedListItem);
+        }
```
timakro: Seems like something that could better go on the input component like

```html
<NumberInputComponent onfocusin={() => focusedItemIndex = myIndex}>
```

If you add a capture phase event handler `onfocusincapture` on the input components like I suggested it would go in there I suppose.

timakro (commented):
Good job on getting rid of two props on the input components and moving some of the logic to search bar. I think we could also move the remaining search bar code out of the input components by using capture phase event handlers. See my comments on the code for the detailed suggestions.

patrick-skowronek (approved):
Except some of the comments of @timakro the rest is fine

timakro (approved):

patrick-skowronek (approved):

PR #708, opened 2025-10-22
chore: make a copy in setCatalogue and setOptions

timakro:
Prevents:
* us from removing stuff from the apps object
* the app from internally mutating the options after setOptions

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -1,6 +1,5 @@
 [book]
 language = "en"
-multilingual = false
 src = "src"
 title = "Lens Book"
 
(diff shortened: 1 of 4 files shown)


patrick-skowronek (changes requested):
I see why this would make sense, but we would override the current options/catalogue regardless of the correctness.

timakro:
Already before this PR the validation was only showing a warning and not stopping the options/catalogue from being loaded. At some point in the past we made this change from error to warning because it was annoying during development to not be able to test stuff just because the schema doesn't 100% match.

patrick-skowronek (approved):

patrick-skowronek (changes requested):
The Documentation should mention that it will be copied. In the case of the original is mutated on propose

patrick-skowronek:
Please also add this to the query store

patrick-skowronek (approved):

PR #714, opened 2025-11-14
Remove system

### LINKED ISSUE #690: Remove "system" from AST
Focus doesn't use it currently and I think it is good this way.

timakro:
Closes #690


patrick-skowronek (approved):

MatsJohansen87:
Didn't we want to make system more generic, instead of removing it completely? As a parameter pass through from catalogue to backend. We could make it be an object to have it flexible.
And if i remember correctly, @patrickskowronekdkfz uses it in one or two projects.

patrick-skowronek:
> Didn't we want to make system more generic, instead of removing it completely? As a parameter pass through from catalogue to backend. We could make it be an object to have it flexible. And if i remember correctly, @patrickskowronekdkfz uses it in one or two projects.

Yes, but as mentioned in #690 I thought about it and did the more work but correct method, If no one else uses it I'm for removing. I agree if we need something like additional information where the key is not sufficient enough, then we could add it, but I think it will not be called system.

book/src/releases/v0.6.3.md
```diff
@@ -0,0 +1,11 @@
+# Version 0.6.3
+
+## Breaking changes
+
+- Removed unused `system` key from catalogue, query store, and AST
```
patrick-skowronek, line 5: It is not a breaking change. Maybe just a "Change"

patrick-skowronek (changes requested):

patrick-skowronek (approved):

PR #721, opened 2025-11-19
feat: sorting charts by buttons

### LINKED ISSUE #303: [Feature]: UI - Make Barcharts sortable after value or alphabetically
### General Summary

The user should have the option to sort bar charts by either by value or alphabetically. I think something like a small toggle switch in one of the corners would be good.

I'm not sure what should be the default for one of them, maybe alphabetically. There should be the option to add your own sort algorithm as it is right now

### Is this Breaking?

No

patrick-skowronek:
### Description

Add sorting buttons for charts. Sort defaults to alpha asc and buttons are visible 

closes #303


src/components/results/ChartComponent.wc.svelte
```diff
@@ -103,6 +105,9 @@
 
     let chart: Chart;
 
+    let sortBy: string = $state("alpha"); // 'value' or 'alpha'
```
timakro: You could restrict this type to `"value" | "alpha"`

src/components/results/ChartComponent.wc.svelte
```diff
@@ -625,6 +683,92 @@
...
+                    fill="none"
+                    stroke="currentColor"
+                    stroke-width="2"
+                >
+                    {#if sortBy === "value" && sortOrder === "asc"}
+                        <!-- Arrow up for ascending -->
+                        <line x1="12" y1="19" x2="12" y2="5"></line>
+                        <polyline points="5 12 12 5 19 12"></polyline>
+                    {:else}
+                        <!-- Arrow down for descending (default) -->
```
timakro: I would change the grayed out button to an up arrow because ascending is what you get when you click the button.

src/components/results/ChartComponent.wc.svelte
```diff
@@ -103,6 +105,9 @@
 
     let chart: Chart;
 
+    let sortBy: string = $state("alpha"); // 'value' or 'alpha'
+    let sortOrder: "asc" | "desc" = $state("asc"); // 'asc' or 'desc'
```
MatsJohansen87, line 109: comment can be removed with type definition

src/components/results/ChartComponent.wc.svelte
```diff
@@ -691,4 +833,42 @@
...
+        border-color: var(--dark-gray, #999);
+    }
+
+    [part~="lens-chart-sort-button"].active {
+        background-color: var(--primary-color, #007bff);
+        border-color: var(--primary-color, #007bff);
+        color: white;
+    }
+
+    [part~="lens-chart-sort-button"]:focus {
```
MatsJohansen87: I would get rid of the outline completely. It isn't aligned with the button styling throughout the library and the indication that this particular button is active is represented by the blue color.

src/components/results/ChartComponent.wc.svelte
```diff
@@ -691,4 +833,42 @@
         top: 0;
         right: 0;
     }
+
+    [part~="lens-chart-sort-buttons"] {
+        position: absolute;
+        top: 0;
+        right: 25px; /* Adjust based on info button width */
+        display: flex;
+        gap: 4px;
```
MatsJohansen87: i would use `gap: var(--gap-xxs);` for consistency.

src/components/results/ChartComponent.wc.svelte
```diff
@@ -691,4 +833,42 @@
...
+        z-index: 1;
+    }
+
+    [part~="lens-chart-sort-button"] {
+        background: transparent;
+        border: 1px solid var(--gray, #ccc);
+        border-radius: 4px;
```
MatsJohansen87: also here better use the variable `border-radius: var(--border-radius-small);`

src/components/results/ChartComponent.wc.svelte
patrick-skowronek: I had it first, but did run into an issue. Is fixed now

src/components/results/ChartComponent.wc.svelte
MatsJohansen87: Small misunderstanding. the border on the button is fine.
i would completely remove this:
```
    [part~="lens-chart-sort-button"]:focus {
        outline: 2px solid var(--focus-color, #007bff);
        outline-offset: 2px;
    }
```

timakro:
Switching between value and alpha sort is still a little unintuitive. It maintains the same sort direction as previous sort but the grayed out button always is a down arrow.

I think grayed out button should always be an up arrow (ascending) and when you click it sort direction should always reset to ascending.

MatsJohansen87:
just saw, that on (even bigger) laptop sizes the buttons are overlapping the heading.

Screenshot 2025-11-21 at 11 03 23

maybe this should be done by flex containers instead of absolute positioning. something like this at a specific breakpoint:
Screenshot 2025-11-21 at 11 14 48

timakro (approved):
Great feature :)

PR #724, opened 2025-11-26
Add indicateApproximation prop to result table and result summary

timakro:
Here example in BBMRI sample locator:

image

Do you think it is worth doing for the result summary or is it sufficient if we add the symbol only to the result table?


MatsJohansen87 (approved):

patrick-skowronek (changes requested):
Code looks fine, documentation is missing

patrick-skowronek (approved):

Issue #497, opened 2025-07-10
[Feature]: Encode selected sites(datarequeststore) into url


patrick-skowronek:
### General Summary

The selected sites for data requests should be included in the URL. This way, when a user returns after making a data request, their previously selected sites remain preselected.

### Is this Breaking?

No

PR #725, opened 2025-11-27
feat: add methode for datarequeststore

patrick-skowronek:
### Description

Adds a site to the datarequeststore

### DIFF
--- a/book/src/releases/v0.6.3.md
+++ b/book/src/releases/v0.6.3.md
@@ -3,6 +3,7 @@
 ## Features
 
 - Charts now support sorting by value or alphabetically with visual toggle buttons that indicate sort direction. Enable this feature by setting the `enableSorting` prop to `true`. Defaults to alphabetical ascending sort.
+- We added functions to remove and add sites to `dataRequestStore`
 
 ## Breaking changes
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -30,8 +30,8 @@ export {
 } from "./helpers/cql-measure";
 export {
     getSelectedSites,
-    setSiteAsSelected,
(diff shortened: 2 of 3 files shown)


timakro:
Are you planning to use this to keep selected sites in sync with the URL? I'd add this functionality to Lens and enable it by default, just as the query is currently synced to the URL.

Not opposed to this PR, though. I would just tend to have the URL syncing logic in Lens. But the function added by this PR can still be useful if an app for whatever reason wants to select/unselect a site.

timakro (commented):
Just looked more closely at the code and saw that you added a new function that is not exposed in index.ts nor called, so something is missing?

patrick-skowronek:
> Are you planning to use this to keep selected sites in sync with the URL? I'd add this functionality to Lens and enable it by default, just as the query is currently synced to the URL.
> 
> Not opposed to this PR, though. I would just tend to have the URL syncing logic in Lens. But the function added by this PR can still be useful if an app for whatever reason wants to select/unselect a site.

We are still planning to do that see #497 and I think this would be also a great feature to have. I have a use case where a project doesn't use the result table and after a query with a result we would set the local site into the datarequeststore. For now we just need the function. After thinking I need a delete function as well since get will not modify the store.

patrick-skowronek:
> Just looked more closely at the code and saw that you added a new function that is not exposed in index.ts nor called, so something is missing?

Good catch, will add the index.ts in a sec

timakro:
Maybe rename them to `selectSite`, `unselectSite`? Consider that users of lens probably don't even know what the DataRequestStore is (not the most intuitive name)

timakro (approved):

PR #726, opened 2025-11-27
Release preparations

timakro:


timakro:
I don't know when this stupid npm bug started biting us, no time to fix it now

PR #734, opened 2025-12-15
fix: map index of data elemete to color mod index

### LINKED ISSUE #728: Fix custom chart colors in combination with value sort
Image

patrick-skowronek:
### Description

fixes #728 

This PR fixes the issue with sort by data and custom colors or when there are more data elements  then colors.

### DIFF
--- a/demo.svelte
+++ b/demo.svelte
@@ -14,7 +14,7 @@
     } from "./src/index";
 
     const barChartBackgroundColors: string[] = ["#052c65", "#0d6efd"];
-    const barChartHoverColors: string[] = ["#344256"];
+    const barChartHoverColors: string[] = ["#000000"];
 
     setOptions({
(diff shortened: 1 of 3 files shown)


patrick-skowronek:
@timakro also there is somewhat of a breaking change. I removed the lines that check if it is a string with comma seperated values. I believe we are now in a position where we don't need that check anymore. What are your thoughts?

timakro (commented):
Looking good. I think backgroundHoverColor is not working (maybe never worked?) Should we remove it?

patrick-skowronek:
> Looking good. I think backgroundHoverColor is not working (maybe never worked?) Should we remove it?

Yes, it was broken by renaming the interface. I fixed it and would still provide it.

timakro (approved):

PR #736, opened 2025-12-16
chore: change default for chart sorting to false

### LINKED ISSUE #727: Improve chart sort button layout

patrick-skowronek:
### Description

This PR changes the default of the sorting of the charts to false

@timakro I tried to fix #727 but the however, I tried to fix the space between the title and the buttons, it always breaks. I wasted enough time with this. If you want you can try it, but I'm ok how it looks and feels right now. I would suggest to developers who need mobile responsiveness to setup there charts with a min-width.


timakro:
@patrick-skowronek I pushed a commit. What do you think?

timakro:
Can you check if docs need to be adjusted now that sort buttons default to disabled?

Maybe we should also mention in the migration guide that backgroundHoverColor was renamed to hoverBackgroundColor

patrick-skowronek:
> Can you check if docs need to be adjusted now that sort buttons default to disabled?
> 
> Maybe we should also mention in the migration guide that backgroundHoverColor was renamed to hoverBackgroundColor

It does look better now, I think the align right makes then more sense.

I will update the documentation and release guide, I stashed them earlier 🤞

timakro (approved):

Issue #745, opened 2026-01-12
[Feature]: Add getSiteResult


patrick-skowronek:
### General Summary

The Lib should have a function that returns a lens result for a given site. 

### Is this Breaking?

No

PR #757, opened 2026-02-03
Add get site result functionality

MatsJohansen87:
### Description

Add function to get the result from a single site

### Related Issue

Solves https://github.com/samply/lens/issues/745

### DIFF
--- a/book/src/guide/results.md
+++ b/book/src/guide/results.md
@@ -85,7 +85,7 @@ if (isLensResultEmpty(result)) {
 Use this function if you want to get the data from a single site.
 
 ```ts
-getSiteResultById("your-site-id");
+const siteResult = getSiteResultById("your-site-id");
 ```
 
 ## Components
--- a/book/src/guide/results.md
+++ b/book/src/guide/results.md
@@ -80,6 +80,14 @@ if (isLensResultEmpty(result)) {
 }
 ```
 
+### Getting values of a single site
+
+Use this function if you want to get the data from a single site.
+
+```ts
+getSiteResultById("your-site-id");
+```
+
 ## Components
 
 Lens provides components to render results.
--- a/book/src/guide/results.md
+++ b/book/src/guide/results.md
@@ -85,7 +85,7 @@ if (isLensResultEmpty(result)) {
 Use this function if you want to get the data from a single site.
 
 ```ts
-const siteResult = getSiteResultById("your-site-id");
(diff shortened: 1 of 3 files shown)


book/src/guide/results.md
```diff
@@ -80,6 +80,14 @@ if (isLensResultEmpty(result)) {
 }
 ```
 
+### Getting values of a single site
+
+Use this function if you want to get the data from a single site.
+
+```ts
+getSiteResultById("your-site-id");
```
timakro:
```suggestion
const siteResult = getSiteResultById("your-site-id");
```

src/stores/response.ts
```diff
@@ -42,6 +42,24 @@ export type LensResult = {
...
 
+/**
+ * Call this when you want to get the result for a specific site.
+ * @param site id of the site to get results
+ * @returns result of the requested site or undefined
+ */
+export function getSiteResultById(site: string): LensResult | undefined {
+    let result: LensResult | undefined;
+    siteResults.subscribe((results) => {
+        console.log("allResults: ", results);
```
timakro:
```suggestion
```

src/stores/response.ts
```diff
@@ -42,6 +42,24 @@ export type LensResult = {
...
+        result = results.get(site);
+    });
+    if (result === undefined || result === null) {
+        console.error(`No result found for site: ${site}`);
```
timakro: I'd also remove this log line - the caller is forced to handle `undefined` type thus should be aware that the operation can fail.

src/stores/response.ts
```diff
@@ -42,6 +42,24 @@ export type LensResult = {
...
 };
 
+/**
+ * Call this when you want to get the result for a specific site.
+ * @param site id of the site to get results
+ * @returns result of the requested site or undefined
+ */
+export function getSiteResultById(site: string): LensResult | undefined {
+    let result: LensResult | undefined;
+    siteResults.subscribe((results) => {
```
timakro: Don't subscribe - subscribe is for when you want to get updates.

timakro (changes requested):
Feels like it would be better named `getSiteResult` to be analogous to `setSiteResult`.

timakro (approved):

timakro:
Thanks!

patrick-skowronek (approved):

Issue #743, opened 2026-01-12
Prevent query and show warning when there are more than one search bar and at least one search bar is empty


timakro:
Kurt Majcen said:

> OR in a search with empty condition/s? does not produce any results, which the user may not understand, as a search with one condition being. empty shows all patients/specimens for all sites.

I agree, when there are more than one search bar and at least one of them is empty we should show a warning to the user (e.g. toast) and not send a query.

patrick-skowronek:
I think we had such a prevention and disabled the search button. I'm not quite sure what happened with that 🤔

timakro:
We could also automatically delete empty search bars when the user clicks the search button.

MatsJohansen87:
> We could also automatically delete empty search bars when the user clicks the search button.

I think that's the best UX. As a user i don't want to take the blame for something, that can be easily handled by the application.

PR #758, opened 2026-02-04
feat: remove empty search bars on search

MatsJohansen87:
### Description
Empty search bars should be deleted when the search is triggered to prevent the AST to have an empty OR
### Related Issue

https://github.com/samply/lens/issues/743


timakro (commented):
Did you test it with only one search bar that is empty?

MatsJohansen87:
> Did you test it with only one search bar that is empty?

fixed 🤦

timakro (approved):
Was too lazy to test myself but looking good:)

PR #775, opened 2026-02-26
Added URL parameter datarequests, it changes based on selected sites

Apocalypse-jk:
### Related Issue

https://github.com/samply/lens/issues/497

---

### How Has This Been Tested?

selecting sites -> copy URL -> select other sites -> paste URL -> press Search and compare selected sites

### DIFF
--- a/src/components/results/ResultTableComponent.wc.svelte
+++ b/src/components/results/ResultTableComponent.wc.svelte
@@ -18,7 +18,6 @@
     import Tooltip from "../informational/Tooltip.svelte";
     import { translate } from "../../helpers/translations";
     import { onMount } from "svelte";
-    import { showToast } from "../../stores/toasts";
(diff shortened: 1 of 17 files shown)


src/components/results/ResultTableComponent.wc.svelte
```diff
@@ -132,6 +136,57 @@
...
+                        ),
+                    ),
+                );
+                datarequestsStore.set(datarequests);
+            } catch {
+                console.error(
+                    "Failed to parse datarequests from URL:",
+                    encodedDatarequests,
+                );
+                showToast(
```
patrick-skowronek: Please don't do a ErrorToast here, in the case of a site is not mappable it should be silent.

patrick-skowronek (changes requested):

patrick-skowronek (approved):

PR #792, opened 2026-03-18
New AST, catalogue, and query types

timakro:
More PRs following:
- Tailwind CSS
- VitePress
- Dropdown list refactor which I initially did when adding the red minus buttons. We'll have to decide if we want to have the red minus buttons. Even if we decide against the red minus buttons this is still a good change.
- Demo with more realistic content (MII Kerndatensatz)

I have the domains changeset locally but would postpone this for now since in my demo it was still confusing for the user. Adding it to the catalogue is a very simple change we can always redo.


.vscode/extensions.json
```diff
@@ -1,3 +1,7 @@
 {
-    "recommendations": ["svelte.svelte-vscode"]
+    "recommendations": [
```
patrick-skowronek, line 2: Do we really want to have the .vscode in the repo? I get the easy start for vs code users

scripts/migrate-catalogue.ts
```diff
@@ -0,0 +1,417 @@
+/**
+ * Converts an old-format Lens catalogue JSON to the new format.
```
patrick-skowronek: I think it is good to have a migration script. It sounds bad but we would need to give them some kind of version. I would propose LensCatalog v0.5 to v0.6 as a name

src/types/catalogue.ts
```diff
@@ -1,172 +1,128 @@
...
-export type Catalogue = Category[];
+type BaseElement = {
+    /** A key that uniquely identifies the catalogue element. */
+    key: string;
+    /** The element's user-facing display name. */
+    name: string;
+    /** Optional text accessed by clicking a "ⓘ" button next to the display name. */
+    infoButtonText?: string[];
+    /** Optional hyperlink shown next to the display name. */
+    infoLink?: {
```
patrick-skowronek, line 13: Do we need this for other project expect ccp?

src/types/query.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type DateRangeItem = BaseItem & {
+    type: "DateRangeItem";
+    /**
+     * Inclusive lower bound in YYYY-MM-DD format.
+     * @format date
+     */
+    min?: string;
+    /**
+     * Inclusive upper bound in YYYY-MM-DD format.
```
patrick-skowronek, line 54: I know this is a far topic, but maybe we should also think about adding time compatibility? In discussion #228 we agreed on n [RFC 3339, section 5.6](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6)

src/index.ts
```diff
@@ -12,7 +12,7 @@ export {
     setQueryStore,
     getQueryStore,
 } from "./stores/query";
-export { setQueryStoreFromAst } from "./helpers/ast-to-query";
+export { convertToLegacyAst } from "./helpers/legacy-ast";
```
patrick-skowronek: This has to go. We agreed to remove this functionality, I think already making it deprecated. Ovis should use the query instead

patrick-skowronek (changes requested):
Part 1/?

I have some questions or small things I would like to discuss

.vscode/extensions.json
timakro: Doesn't have any disadvantages for users of other IDEs. Would add it unless you think one of these extensions are not useful for VS Code users.

src/types/catalogue.ts
timakro: Only Test DKTK Explorer is using this right now. I think we should remove `infoButtonText` and `infoLink` and instead add `infoText: string` which takes markdown. This PR would be a good opportunity to do this and also convert old catalogues.

scripts/migrate-catalogue.ts
timakro: Good point -- I updated the comment and renamed the script to clarify

src/types/query.ts
timakro: I think contrary to what we decided back then time should be optional in the AST. My suggestion would be to change the type to `min: { date: string }?` now, so we have the freedom to add an optional `time` field later.

src/index.ts
timakro: I made two changes here:
* Removed `setQueryStoreFromAst` only used by Ovis -- I think this is the right move, see reasons in the RFC
* Exported `convertToLegacyAst` -- We discussed not exporting this and instead copying it into the applications. Instead I suggest we export it from Lens but already mark it deprecated.

scripts/migrate-catalogue.ts
patrick-skowronek: 👍 I guess we are skipping version 0.6 😅

src/types/catalogue.ts
patrick-skowronek: Good Idea, maybe move it to another PR?

patrick-skowronek (approved):

PR #829, opened 2026-05-27
playwright tests

paolaklein:
### Description

This PR introduces end-to-end testing with Playwright, adding specs that cover the core UI workflows including search bar autocomplete, query building (AND/OR logic), result display, catalogue interaction, language switching, and toast notifications.
Playwright is wired into both the standard CI linting pipeline and the release verification workflow, with support for running against either a local dev server or the live GitHub Pages demo.


.github/workflows/verify-new-release.yml
```diff
@@ -20,3 +20,12 @@ jobs:
                     echo "::error:: Please run `npm run version` before merging to main!"; \
                     exit 1; \
                   fi
+            - uses: actions/setup-node@v6
+              with:
+                  node-version: latest
+            - run: npm ci
+            - run: npx playwright install chromium --with-deps
```
patrick-skowronek, line 27: This should be using --only-shell to only install headless

.github/workflows/linting.yml
```diff
@@ -22,6 +22,10 @@ jobs:
             - run: npx eslint .
             - run: npx vite build
             - run: npx vitest
+            - run: npx vite build --config vite.config.demo.js
+            - run: npx playwright install chromium --with-deps
```
patrick-skowronek, line 26: This should be using --only-shell to only install headless

patrick-skowronek, line 26: Should the default be chrome? Maybe it would make sense for pull request to also test in firefox and webkit(safari)

src/helpers/ast-to-query.test.ts
```diff
@@ -181,3 +181,66 @@ test("setQueryStoreFromAst: string value", () => {
         ],
     });
 });
+
+test("getAst: empty queryStore returns top-level OR with no children", () => {
```
patrick-skowronek, line 186: We should document what is the default for an empty query store. For me an empty or is fine. But in the future where it could start with a "and" we should be prepared.

src/helpers/ast-to-query.test.ts
```diff
@@ -181,3 +181,66 @@ test("setQueryStoreFromAst: string value", () => {
...
+    setQueryStoreFromAst({ operand: "OR", children: [] });
```
patrick-skowronek: This function is deprecated and should be tested. Instead please test the QueryStore2AST

src/helpers/human-readable.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    test("returns empty string for an empty query", () => {
+        expect(getHumanReadableQueryAsFormattedString()).toBe("");
+    });
+
+    test("1-group query contains the group header and the filter", () => {
+        setQueryStore([
+            [makeItem("First name", [{ name: "Olaf", value: "Olaf" }])],
+        ]);
+        const result = getHumanReadableQueryAsFormattedString();
+        // translate("query_info_header") → "Search ANY of the following groups"
```
patrick-skowronek: Why is this commented out? Testing translate should be done in the translate file.

test-results/.last-run.json
```diff
@@ -0,0 +1,4 @@
+{
+  "status": "passed",
```
patrick-skowronek, line 2: This should not be committed and be set in the .gitignore

tests/e2e/catalogue.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        catalogue.getByText("Diagnosis ICD-10", { exact: true }),
+    ).toBeVisible();
+    await expect(catalogue.getByText("Diagnosis age", { exact: true })).toBeVisible();
+});
+
+test("6.3 clicking Gender 'male' criterion in catalogue adds a chip", async ({
+    page,
+}) => {
+    const catalogue = page.locator("lens-catalogue");
+    // Expand the group first
```
patrick-skowronek: Do we need the comments?

tests/e2e/searchbar-helpers.ts
patrick-skowronek, line 1: Maybe change the file name. This tests mostly the searchbar/querystore functionality.

patrick-skowronek (commented):
The e2e Test look good, but there are some things in the tests. Please add documentation for the e2e test, as a test plan. Also add a section to the lens book what and how changes should be documented.

timakro:
Whoof, adding a giant heap of vibe-coded integration tests just feels like the wrong move to me especially given that Lens must undergo major changes in the near future. This feels like just adding tests onto everything without consideration if this is actually something that is likely to break.

I've just read a few tests (there are too many for me to read though) but the ones I've seen are kind of tightly coupled to the DOM structure. For example the tests for the query explain button seem basically to enforce the exact behavior and DOM structure of how the query explain button renders out the query right now. Any changes to the query explain button would require changing the tests accordingly. And I just can't see what regression this would actually catch.

PR #787, opened 2026-03-11
Added Catalogue Types to Lens Book

Apocalypse-jk:
### Description

Added Catalogue Types to Lens Book

### Related Issue

[#742](https://github.com/samply/lens/issues/742)

PR #831, opened 2026-05-28
feat: added Catalogue Types to Lens Book

patrick-skowronek:
### Description

This PR is based on #787 but based on updated develop.

All credits to @Apocalypse-jk


timakro (approved):
Let's merge this for now but long term I'd prefer a solution that auto generates docs from the type definitions. We already have TypeDoc (e.g. https://samply.github.io/lens/docs/types/NumericRangeCategory.html) but the presentation there is not ideal. Would be worth looking into if it can be configured.

PR #833, opened 2026-05-29
Feature/add readonly option to searchbar

### LINKED ISSUE #830: [Feature]: Readonly option for Searchbar
### General Summary

Add a "readonly" prop to the searchbar component, which deactivates any interactive functionality of the searchbar except for the info buttons.

### Is this Breaking?

No

thkoehler11:
PR to Issue #830 

Adds the option to display the Searchbar in readOnly mode, which negates any interactive behavior.

### DIFF
--- a/book/src/components/searchbar.md
+++ b/book/src/components/searchbar.md
@@ -32,13 +32,13 @@ The `lens-searchbar` component offers an interface for exploring of all catalogu
 
 ## Props
 
-| Prop                    | Type      | Default                                       | Description                                                       |
-|-------------------------|-----------|-----------------------------------------------|-------------------------------------------------------------------|
-| `noMatchesFoundMessage` | `string`  | `"No matches found"`                          | Message shown when no autocomplete options are found.             |
-| `typeMoreMessage`       | `string`  | `"Search will start with 3 inserted letters"` | Message shown when input is too short for autocomplete.           |
-| `placeholderText`       | `string`  | `"Type to filter conditions"`                 | Placeholder in the search input field.                            |
-| `index`                 | `number`  | `0`                                           | Used to manage multiple search bars (e.g., in groups or filters). |
-| `readOnly`              | `boolean` | `false`                                       |  Makes the component read-only from the user's perspective. Values can still be modified programmatically via the store. |
+| Prop                    | Type      | Default                                       | Description                                                                                                             |
+| ----------------------- | --------- | --------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


patrick-skowronek:
Hey @thkoehler11, thanks for the PR. Can you us something about the use case of a read only searchbar? Is it still intended to add items trough diagrams or catalogues?

thkoehler11:
> Hey @thkoehler11, thanks for the PR. Can you us something about the use case of a read only searchbar? Is it still intended to add items trough diagrams or catalogues?

Hi, we want to integrate the Searchbar into the "Samply Requester" / "Project Manager" and its supposed just for the visual representation of the query. So no items will be added or deleted outside of Lens. The idea was, that the user will immediately recognize the query, when he's forwarded from Lens to PR.

book/src/components/searchbar.md
```diff
@@ -32,12 +32,13 @@ The `lens-searchbar` component offers an interface for exploring of all catalogu
...
+| Prop                    | Type      | Default                                       | Description                                                       |
+|-------------------------|-----------|-----------------------------------------------|-------------------------------------------------------------------|
+| `noMatchesFoundMessage` | `string`  | `"No matches found"`                          | Message shown when no autocomplete options are found.             |
+| `typeMoreMessage`       | `string`  | `"Search will start with 3 inserted letters"` | Message shown when input is too short for autocomplete.           |
+| `placeholderText`       | `string`  | `"Type to filter conditions"`                 | Placeholder in the search input field.                            |
+| `index`                 | `number`  | `0`                                           | Used to manage multiple search bars (e.g., in groups or filters). |
+| `readOnly`              | `boolean` | `false`                                       | Deactivates any interactive behavior, except the Info button      |
```
patrick-skowronek: I don't like the description of this prop. In my opinion it should be stated that you could add elements to the bar regardless

patrick-skowronek:
The implementation is fine. Just as a question, would the query explain button do the same Job?

book/src/components/searchbar.md
thkoehler11: How about "Deactivates any interactive behavior of the Searchbar itself (except the Info Buttons), changes to the query by other components like Catalogue will still update the query." ?

patrick-skowronek:
```suggestion
| `readOnly`              | `boolean` | `false`                                       |  Makes the component read-only from the user's perspective. Values can still be modified programmatically via the store. |
```

patrick-skowronek (approved):

